PR #1, opened 2019-10-07
GBJ: GP19-38

GBurley-Jones:


dex-mendeley:
I see you have commited a few files that you probably don't need. You've added .DS_STORE files that are specific to MacOS, so it might make sense to remove those and add the rule to the .gitignore file. 

Adding the .mvn folder and mvnw file usually divides developers, some like to include it, other don't. So it's probably fine to keep :)

H-Ford:
> I see you have commited a few files that you probably don't need. You've added .DS_STORE files that are specific to MacOS, so it might make sense to remove those and add the rule to the .gitignore file.
> 
> Adding the .mvn folder and mvnw file usually divides developers, some like to include it, other don't. So it's probably fine to keep :)

We've got rid of the maven wrapper instead just downloading maven to run the project 😃

dex-mendeley (approved):
Looks good to me now :)

H-Ford (approved):
Looks good, make sure you have all downloaded maven (brew install maven)

singhals-els (approved):

PR #3, opened 2019-10-08
GP19-28_Design_Database_Schema

karenzamani:
This branch creates a conference object for ticket GP19-28.


SpringProject/src/main/resources/db.migration/V1__createSchema.sql
```diff
@@ -0,0 +1,8 @@
+create schema conferenceFinder;
```
Piotrek605: I would keep all your SQL files consistent, so CREATE SCHEMA and CREATE TABLE should be in block capitals because that's the convention and that's what the second SQL file looks like.

SpringProject/src/test/java/com/GradProject2019/SpringProjectApplicationTests.java
```diff
@@ -1,4 +1,4 @@
-package com.GradProject2019.SpringProject;
+package com.GradProject2019;
```
dex-mendeley: I assume you want both the SpringBootApplicationTests and SpringBootApplication files to be in the same package, so you'd probably want to rename this to `package com.gradproject2019`, too :)

SpringProject/src/main/resources/db.migration/V2__addUsers.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TABLE USERS (ID INT AUTO_INCREMENT PRIMARY KEY, USERID VARCHAR(45));
```
dex-mendeley: Should this table not also be added to the schema?

SpringProject/src/main/resources/db.migration/V1__createSchema.sql
```diff
@@ -0,0 +1,8 @@
...
+create table conferenceFinder.conference (
+	id long,
```
Piotrek605: Wrong indentation here, so this line should be at the same level as the lines below.

SpringProject/src/main/resources/db.migration/V2__addUsers.sql
Piotrek605: This should be 'user', 'id' and 'user_id', all lower case, right @H-Ford @vsocolov-elsevier @mdykiert ?
What's the difference between id and user_id?

SpringProject/src/test/java/com/GradProject2019/SpringProjectApplicationTests.java
Piotrek605: Package names should start with a lower-case letter.

Piotrek605 (changes requested):

SpringProject/src/main/resources/db/migration/V1__create_schema.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE conference (
+	id BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
```
Piotrek605: The indentation is still off here.

SpringProject/src/main/resources/db/migration/V2__add_test_data.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO conference(conference_name, conference_date_time, city, conference_description, conference_topic)
```
Piotrek605: I don't think test data is something you should add as a migration, migrations should only set up tables without adding any data. Later on you might have some automation in place so that when you run tests, you automatically insert test data and clear it afterwards but for now I would only add data manually (by running this SQL in the workbench).

Piotrek605 (changes requested):

SpringProject/pom.xml
```diff
@@ -24,6 +26,25 @@
 			<artifactId>spring-boot-starter</artifactId>
 		</dependency>
 
+<!--		<dependency>-->
+<!--			<groupId>org.flywaydb</groupId>-->
+<!--			<artifactId>flyway-core</artifactId>-->
+<!--		</dependency>-->
```
H-Ford: Remove this dependency if it is not being used

GBurley-Jones: In our version this is not commented out, do you think we need to remove it?

Piotrek605 (approved):

SpringProject/src/main/resources/application.yml
```diff
@@ -0,0 +1,16 @@
...
+  application:
+    name: SpringProject
+  datasource:
+    platform: mysql
+    driverClassName: com.mysql.cj.jdbc.Driver
+    url: jdbc:mysql://localhost:3306/conference_finder?autoreconnect=true
+    password: password
+    username: root
+
+#flyway:
```
dex-mendeley: The same goes for this: If you're not using it, it's better to not comment it out but just to remove it ;)

dex-mendeley (changes requested):

dex-mendeley (approved):

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,28 @@
+<component name="ProjectCodeStyleConfiguration">
```
dex-mendeley: You probably shouldn't commit these files, they're specific to IntelliJ ;)

.idea/grads19-be.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dex-mendeley: As my previous comment, you probably want to remove this file as it's specific to IntelliJ

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dex-mendeley: Same

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dex-mendeley: Again ;)

.idea/workspace.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dex-mendeley: Here, too :P

dex-mendeley (changes requested):

PR #4, opened 2019-10-10
Addition of repo layer

karenzamani:
We added in a repo java class and an implement repo java class as part of ticket GP19-32


SpringProject/src/main/java/ConferenceRepository.java
```diff
@@ -0,0 +1,4 @@
+package PACKAGE_NAME;
+
+public class ConferenceRepository {
```
dex-mendeley: I assume this will become an interface extending a JpaRepository/CRUDRepository?

SpringProject/src/main/java/ConferenceRepoImp.java
```diff
@@ -0,0 +1,4 @@
+package PACKAGE_NAME;
```
dex-mendeley: Lovely package name but maybe not the best choice here :P

dex-mendeley (changes requested):

dex-mendeley:
You guys sure this is ready for a PR yet?

SpringProject/src/main/java/com/gradproject2019/ConferenceRepository.java
```diff
@@ -0,0 +1,12 @@
+package com.gradproject2019;
+
+import java.util.List;
```
H-Ford: Unused import. Control + Option + o to optomise imports

SpringProject/src/main/java/com/gradproject2019/ConferenceRepository.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.data.repository.CrudRepository;
+
+public interface ConferenceRepository extends CrudRepository <Conference, Long>{
+
```
H-Ford: Delete lines

SpringProject/src/main/java/com/gradproject2019/Conference.java
```diff
@@ -1,14 +1,27 @@
...
+
+import javax.persistence.*;
 import java.time.Instant;
 
+@Entity
 public class Conference {
-    private long id;
+    @Id
+    private Long id;
+    @Column(name = "conference_name")
```
H-Ford: Do the column names need to have the conference prefix if they are part of the conference object and conference table?

SpringProject/src/main/java/com/gradproject2019/ConferenceRepository.java
```diff
@@ -0,0 +1,12 @@
+package com.gradproject2019;
+
+import java.util.List;
+
+import org.springframework.data.repository.CrudRepository;
+
+public interface ConferenceRepository extends CrudRepository <Conference, Long>{
```
singhals-els: nit: The formatting of the class seems a little off, too much white space. Option + Command + L

SpringProject/src/main/java/com/gradproject2019/Conference.java
```diff
@@ -1,14 +1,27 @@
+package com.gradproject2019;
+
+import javax.persistence.*;
 import java.time.Instant;
 
+@Entity
```
H-Ford, line 6: Put the table name in here as "conferences" otherwise it will generate as "conference", I have a look in your db migration V1 and you will also need to change the table name to conferences there. https://stackoverflow.com/questions/29087626/entity-class-name-is-transformed-into-sql-table-name-with-underscores

H-Ford (changes requested):

SpringProject/src/main/java/com/gradproject2019/Conference.java
GBurley-Jones: Name, DateTime and Description are pre-determined types in SQL therefore it was our understanding that we could not use them as column names, this is why we re-named them without the prefix "conference" within the class

singhals-els: I think it is fine to have the prefix as if you add more tables later requiring the column "name" they are easier to differentiate from each other.

SpringProject/src/main/java/com/gradproject2019/Conference.java
GBurley-Jones: We are a bit unclear on this, should we rename the Conference class?

singhals-els: the Conference class will remain singular

SpringProject/src/main/java/com/gradproject2019/Conference.java
Piotrek605: Did you try it? I've never heard of 'name' and 'description' being reserved keywords in SQL.

Piotrek605 (approved):

singhals-els (approved):

H-Ford (approved):

SpringProject/src/main/java/com/gradproject2019/Conference.java
```diff
@@ -1,14 +1,27 @@
+package com.gradproject2019;
+
+import javax.persistence.*;
```
rochinae, line 3: avoid * imports

SpringProject/src/main/java/com/gradproject2019/Application.java
```diff
@@ -1,13 +1,38 @@
...
 	public static void main(String[] args) {
 		SpringApplication.run(Application.class, args);
 	}
 
-}
+	@Transactional(readOnly = true)
+	@Override
+	public void run(String... args) throws Exception {
+
+		System.out.println("DATASOURCE = " + dataSource);
```
rochinae, line 27: use logging (sfl4j I guess...)

PR #5, opened 2019-10-11
created controller layer

karenzamani:


singhals-els (changes requested):
This is not very RESTful. I would suggest calling the path just "conferences". The GET endpoint indicates to the user the transaction that is being performed. No need for the ListOf.

SpringProject/src/main/java/com/gradproject2019/Controller.java
```diff
@@ -1,13 +1,20 @@
...
 @org.springframework.stereotype.Controller
-@RequestMapping(path="/all")
+//@RequestMapping(path="/all")
 public class Controller {
 
     @Autowired
     private ConferenceRepository conferenceRepository;
 
-}
+    @GetMapping(path="/ListOfConferences")
```
dex-mendeley: I don't think that's a very good path name. REST suggests paths that describe resources, so maybe something like 
/conferences/all or conferences/list.

Then later one you could do something like conferences/conference?id=123 if you're looking for individual conferences

SpringProject/src/main/java/com/gradproject2019/Controller.java
```diff
@@ -1,13 +1,20 @@
 package com.gradproject2019;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+
 
 @org.springframework.stereotype.Controller
```
dex-mendeley: Probably makes sense to import this and use just the simple @Controller annotation

dex-mendeley (changes requested):

H-Ford:
In the future, commits should have the ticket tag as a prefix like your branch name.
For example "GP19-30 Create the controller layer for conferences"
Allows you to keep relate the implementation of a feature to the business requirements

SpringProject/src/main/java/com/gradproject2019/Controller.java
```diff
@@ -1,13 +1,20 @@
...
+//@RequestMapping(path="/all")
 public class Controller {
 
     @Autowired
     private ConferenceRepository conferenceRepository;
 
-}
+    @GetMapping(path="/ListOfConferences")
+    public @ResponseBody Iterable<Conference> getAllUsers(){
+        return conferenceRepository.findAll();
```
H-Ford: You will need a service layer between the controller and the repository layer. You can do that on this branch and do both tickets at once.

H-Ford: (Not best practise but for the sake of expediting things)

SpringProject/src/main/java/com/gradproject2019/Controller.java
singhals-els: tbh I do not think the /all or /list is needed in this case

singhals-els (approved):

PR #7, opened 2019-10-14
GP19-47: create and test find conference by id method

GBurley-Jones:


SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -18,4 +21,12 @@ public ConferenceServiceImpl(ConferenceRepository conferenceRepository) {
         return conferenceRepository.findAll();
     }
 
+    @Override
+    public Conference findConferenceById(Long conferenceId) {
+        Optional<Conference> conference = conferenceRepository.findById(conferenceId);
+        if(conference.isPresent())
+            return conference.get();
+
+        return null;
```
vsocolov-elsevier: Do not return null. Generally it is a good practice to return an Optional for single object and Collection.empty() for list/sets.

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -18,4 +21,12 @@ public ConferenceServiceImpl(ConferenceRepository conferenceRepository) {
         return conferenceRepository.findAll();
     }
 
+    @Override
+    public Conference findConferenceById(Long conferenceId) {
+        Optional<Conference> conference = conferenceRepository.findById(conferenceId);
+        if(conference.isPresent())
```
vsocolov-elsevier: Never use isPresent and get method from Optional. It's a smell and it means that smth is wrong in your method.

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceService.java
```diff
@@ -5,4 +5,6 @@
 public interface ConferenceService {
 
     Iterable<Conference> listConferences();
```
vsocolov-elsevier: Is it a real reason to use Iterable<> instead of using specific type?

SpringProject/src/main/java/com/gradproject2019/conferences/controller/ConferenceController.java
```diff
@@ -14,8 +15,15 @@
...
     @GetMapping(path = "/conferences")
-    public @ResponseBody
-    Iterable<Conference> getAllUsers() {
+    @ResponseBody
+    public Iterable<Conference> getAllUsers() {
         return conferenceService.listConferences();
     }
+
+    @GetMapping(path = "/conferences/{id}")
+    @ResponseBody
```
vsocolov-elsevier: No need for ResponseBody annotation. If your jackson mapping is properly configured in your project it will do conversion by default.

SpringProject/src/main/java/com/gradproject2019/conferences/controller/ConferenceController.java
```diff
@@ -14,8 +15,15 @@
...
+    public Conference getConferenceById(@PathVariable(required = false) Long conferenceId) {
```
vsocolov-elsevier: @PathVariable(required = false), usually path variables are mandatory. If you need them to be optional it's better to use RequestParams

SpringProject/pom.xml
```diff
@@ -59,6 +59,28 @@
...
+
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-all</artifactId>
+			<version>1.9.5</version>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.opentest4j</groupId>
```
vsocolov-elsevier: Is it a specific reason to use opentest4j? Which features from opentest4j are you using in this project?

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
dex-mendeley: I disagree. In this case I think it's perfectly fine to use an Optional to handle possible null values from the repository.

vsocolov-elsevier: Repository layer should return Optional as well. As I already said returning null values is a bad practice. That's why in Java 8 Optionals were introduced.

GBurley-Jones: We have been trying to research this and haven't found an alternative to the isPresent method, please see linked Oracle page:https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html ,  do you have a suggestion of another method? Thank you

vsocolov-elsevier (approved):

PR #9, opened 2019-10-28
Gp 7 create a conference fix

GBurley-Jones:


SpringProject/src/it/java/com/gradproject2019/ConferenceControllerIT.java
```diff
@@ -92,20 +92,20 @@ public void shouldReturn404WhenIdDoesNotExist() throws URISyntaxException {
...
-
-        //when
-        ResponseEntity<ConferenceResponseDto> responseConference = getSingleConference(uri);
-
-        //Then
-        Assert.assertEquals(200, responseConference.getStatusCodeValue());
-        Assert.assertEquals(conference.getId(), responseConference.getBody().getId());
-        Assert.assertEquals(conference.getName(), responseConference.getBody().getName());
-    }
+//    @Test  CANNOT FUNCTION WHILE NOT USING DOCKER (CONTAINERISATION)
```
rochinae: you can also put @Ignore just after @Test. The changes are easier to track as only one line modified

GBurley-Jones: Okay, we will change this now!

SpringProject/src/it/java/com/gradproject2019/ConferenceControllerIT.java
```diff
@@ -92,7 +89,7 @@ public void shouldReturn404WhenIdDoesNotExist() throws URISyntaxException {
         Assert.assertEquals(404, responseConference.getStatusCodeValue());
     }
 
-    @Test
+    @Test  @Ignore //CANNOT FUNCTION WHILE NOT USING DOCKER (CONTAINERISATION)
```
vsocolov-elsevier: Why did you write a test if you cannot test it and it's not working at the moment?
If you've decided to commit this code then it should work, otherwise it's a dead code and should be removed from the project.

Piotrek605 (approved):

PR #14, opened 2019-11-11
Gp19 70 hash passwords

karenzamani:


SpringProject/src/test/java/com/gradproject2019/PasswordServiceTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PasswordServiceTest {
+
+    @InjectMocks
```
vsocolov-elsevier: No need to use Mockito here. You are not using mocks anywhere in this test.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -13,15 +15,26 @@
...
         this.userRepository = userRepository;
+        this.passwordService = passwordService;
     }
 
     @Override
     public UserResponseDto saveUser(UserRequestDto userRequestDto) {
         User user = from(userRequestDto);
-        //Check username and password
+        checkPassword(user.getPassword());
+        user.setPassword(passwordService.hash(userRequestDto.getPassword()));
```
H-Ford: Can extract password as a variable

SpringProject/src/it/java/com/gradproject2019/UserControllerIT.java
```diff
@@ -81,6 +81,19 @@ public void shouldReturn400WhenAnyFieldNull() throws URISyntaxException {
         Assert.assertEquals(400,response.getStatusCodeValue());
     }
 
+    @Test
+    public void shouldReturn400WhenInvalidPassword() throws URISyntaxException {
```
Piotrek605: This name is a bit confusing, I thought it was testing the login. I'd rename it to something like shouldReturn400WhenPasswordWrongFormat or ...PasswordInvalidFormat. It might be helpful to rename anything containing 'invalid password' in the rest of the code too.

Piotrek605 (changes requested):

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -13,15 +15,26 @@
 
     private UserRepository userRepository;
 
-    public UserServiceImpl(UserRepository userRepository) {
+    @Autowired
```
vsocolov-elsevier: You can remove @Autowired from here because your services are injected in constructor. 
Also I would recommend to put your services as final, it's a good practice to make them final when you inject them via constructor.

SpringProject/src/test/java/com/gradproject2019/PasswordServiceTest.java
MerrettM: Removing this results in a null pointer exception?

vsocolov-elsevier: Your Password service doesn't have any dependency, it's a simple util class, you can create it via new, or in general make it static class with utility methods in it. It doesn't have any business logic in it.

MerrettM: In the future this will become part of a new Authorisation Controller, Service, Repository sequence, which is why it's called service and has the `@Service` tag. does that mean we would have to have the mockito stuff?

vsocolov-elsevier: "In the future" is a very abstract thing (it could be tomorrow, but it could be in 2 months time). If you do not need it now, then you have to keep it minimal.

Piotrek605 (approved):

vsocolov-elsevier (approved):

PR #16, opened 2019-11-12
Gp19 79 new auth endpoint

karenzamani:


Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+    private AuthRepository authRepository;
+
+    @Autowired
+    private UserRepository userRepository;
+
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @LocalServerPort
+    int testServerPort = 8080;
```
vsocolov-elsevier: You do not need to hardcode port here, because above you have used SpringBootTest.WebEnvironment.RANDOM_PORT

SpringProject/src/main/java/com/gradproject2019/auth/data/LoginDto.java
```diff
@@ -0,0 +1,53 @@
+package com.gradproject2019.auth.data;
+
+import org.springframework.validation.annotation.Validated;
+
+import javax.validation.constraints.NotNull;
+
+@Validated
+public class LoginDto {
+    @NotNull(message = "Invalid entry in login field.")
```
vsocolov-elsevier: Please extract this message in a constant, no need to duplicate code.

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
MerrettM: How would we assign the port to the `baseUrl` later in the code, line 55 I think?

SpringProject/src/main/resources/db/migration/V3__create_schema.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE tokens (
```
vsocolov-elsevier, line 1: What was the reason behind extracting tokens in a separate table? Do you really need to normalise this table? Remember that each join is a time consuming operation.

MerrettM: I think it's so we can have a separate uuid for each session which are posted and deleted on login and logout respectively

SpringProject/src/main/java/com/gradproject2019/auth/data/LoginDto.java
MerrettM: Would it be better to be more specific about which field is invalid? eg. `Invalid entry in username field`?

SpringProject/src/main/resources/db/migration/V3__create_schema.sql
MerrettM: Harry has explained that the reason we are doing it in the way we are is so that users can be logged in on multiple devices at once.

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+        realPassword = PasswordUtils.hash("P455w0rd!");
+        user = new User(1L, "KaramsCoolUsername", "Karam", "Kapoor", "KSinghK@gmail.com", realPassword, "Botanist");
+        baseUrl = "http://localhost:" + testServerPort + "/auth";
+        restTemplate.getRestTemplate().setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+    }
+
+    @After
+    public void tearDown() {
+        authRepository.deleteAll();
+        userRepository.deleteAll();
```
ZaryabK: If you'll be deleting from multiple repositories before and after every test, this could be extracted out in its own method (maybe statically called from a test utils class?)

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+        ResponseEntity<Token> response = login(uri, request);
+
+        //then
+        Assert.assertEquals(404, response.getStatusCodeValue());
+    }
+
+    @Test
+    public void shouldReturn403nWhenUserExistsAndPasswordIncorrect() throws URISyntaxException {
+        //given
+        URI uri = new URI(baseUrl +"/login");
```
ZaryabK: The same URI is being constructed in every test, so this could be extracted into it's own `static final` field.

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+    private User user;
+    private String realPassword;
+    private String baseUrl;
+
+    @Before
+    public void setUp() {
+        authRepository.deleteAll();
+        userRepository.deleteAll();
+        realPassword = PasswordUtils.hash("P455w0rd!");
+        user = new User(1L, "KaramsCoolUsername", "Karam", "Kapoor", "KSinghK@gmail.com", realPassword, "Botanist");
```
ZaryabK: I hope this isn't a real email address!

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+        User savedUser = userRepository.saveAndFlush(user);
+        ResponseEntity<Token> response = login(uri, request);
+
+        //then
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(savedUser.getId(), response.getBody().getUserId());
+    }
+
+    @Test
+    public void shouldReturn404WhenUserDoesNotExists() throws URISyntaxException{
```
ZaryabK: Whitespace before opening curly brace

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+        URI uri = new URI(baseUrl +"/login");
+        LoginDto loginDto = createLoginDto("KaramsCoolUsername", "WrongPassword");
+        HttpEntity<LoginDto> request = new HttpEntity<>(loginDto);
+
+        //when
+        userRepository.saveAndFlush(user);
+        ResponseEntity<Token> response = login(uri, request);
+
+        //then
+        Assert.assertEquals(403, response.getStatusCodeValue());
```
ZaryabK: This test should also assert on the message returned (if in fact one is being returned to the client, and if this is not the case, then why?)

SpringProject/src/main/java/com/gradproject2019/auth/service/AuthServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+    public AuthServiceImpl(AuthRepository authRepository, UserRepository userRepository) {
+        this.authRepository = authRepository;
+        this.userRepository = userRepository;
+    }
+
+    @Override
+    public Token login(LoginDto loginDto) {
+       User user = userRepository
+               .findByUsername(loginDto.getUsername())
+               .orElseThrow(UserNotFoundException::new);
```
ZaryabK: Nice use of optionals here!

SpringProject/src/main/java/com/gradproject2019/auth/exception/IncorrectPasswordException.java
```diff
@@ -0,0 +1,8 @@
+package com.gradproject2019.auth.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.FORBIDDEN, reason="Password incorrect.")
```
ZaryabK: This may be a pedantic point, but I question the security of your app based off of the exception messages (if these are the messages displayed on the client)

I could try multiple usernames until I get a "password incorrect" message, and from that I would understand that I have found a user of your application. 

You could fix this by changing the message returned by UserNotFound and IncorrectPassword exceptions (maybe something like "Username or password incorrect"?)

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private User user;
+    private String realPassword;
+    private String baseUrl;
+
+    @Before
+    public void setUp() {
+        authRepository.deleteAll();
+        userRepository.deleteAll();
+        realPassword = PasswordUtils.hash("P455w0rd!");
```
H-Ford: realPassword? hashedPassword would be better or even just password

SpringProject/src/main/java/com/gradproject2019/auth/persistance/Token.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "tokens")
+public class Token {
+    @Id
+    private Long userId;
+    private String token;
```
ZaryabK: Just wondering, is this actually a String or is it a UUID?

H-Ford (changes requested):
You're pretty much there, cracking stuff!

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
ZaryabK: This also applies to the user not found test above this test in the same file.

ZaryabK: This also applies to the user not found test above this one.

SpringProject/src/it/java/com/gradproject2019/AuthControllerIT.java
MerrettM: Don't worry it isn't!

SpringProject/src/main/java/com/gradproject2019/auth/controller/AuthController.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Autowired
+    private AuthService authService;
+
+    @PostMapping(path = "/auth/login")
+    public ResponseEntity<Token> login(@Valid @RequestBody LoginDto loginDto) {
+        Token token = authService.login(loginDto);
+        return ResponseEntity.ok(token);
+    }
+
```
H-Ford: Whitespace

SpringProject/src/main/java/com/gradproject2019/users/exception/UserNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package com.gradproject2019.users.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.NOT_FOUND, reason="User not found.")
+public class UserNotFoundException extends RuntimeException {
+
```
H-Ford: Whitespace

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -1,9 +1,17 @@
...
+
+        import java.util.Optional;
 
 @Repository
 public interface UserRepository extends JpaRepository<User, Long> {
-}
+
+    @Query (value = "SELECT * FROM users where username = :username", nativeQuery = true)
+    Optional<User> findByUsername(@Param("username") String username);
+
```
H-Ford: Whitespace

SpringProject/src/main/java/com/gradproject2019/auth/persistance/Token.java
karenzamani: Hi Zabi, it should be a UUID so we've changed the type from String to UUID. Thanks!!

singhals-els (approved):

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -0,0 +1,120 @@
...
+    @Autowired
+    private TestRestTemplate restTemplate;
+
+    @LocalServerPort
+    int testServerPort;
+
+    private User user;
+    private String hashedPassword;
+    private URI loginUri;
+    private String baseUri;
```
ZaryabK, line 39: Pedantic comment: fields of the same type should be together (group all fields of type String together)

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @After
+    public void tearDown() {
+        clearRepositories();
+    }
+
+    @Test
+    public void shouldReturn200andCreateTokenWhenUserExistsAndPasswordCorrect() {
+        //given
+        LoginDto loginDto = createLoginDto("KaramsCoolUsername", "P455w0rd!");
```
ZaryabK: This username and password (that I've noticed you're also using when creating `user`, as well as throughout this file in other tests) should be extracted into their own `final static` fields

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -0,0 +1,120 @@
...
+        ResponseEntity<Token> response = login(loginUri, request);
+
+        //then
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(savedUser.getId(), response.getBody().getUserId());
+        Assert.assertNotNull(response.getBody().getToken());
+    }
+
+    @Test
+    public void shouldReturn401WhenUserDoesNotExists() {
```
ZaryabK: Typo here: "DoesNotExist"

H-Ford (approved):

ZaryabK (changes requested):

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
karenzamani: For the last test we change the password to test what would happen if an incorrect password was added. Shall we extract username and password and then overwrite password in the last test?

KaramKapoor: We have extracted the loginDto and all the tests still work, including the last one which checks for the incorrect password. I'm assuming the only change is that it just overwrites the loginDto password field with "WrongPassword", and uses the same object?

PR #17, opened 2019-11-20
Grads 0 validate username email

KaramKapoor:


SpringProject/src/main/java/com/gradproject2019/users/exception/CredentialsExistException.java
```diff
@@ -0,0 +1,8 @@
+package com.gradproject2019.users.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.CONFLICT, reason="Credentials are already in use.")
+public class CredentialsExistException extends RuntimeException {
```
Piotrek605: I think I would replace 'credentials' with 'userInfo' or 'username or email', otherwise it seems like I can guess a user's username and password by trying to register one with the same credentials and getting this exception.

Piotrek605 (changes requested):

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -15,4 +23,39 @@
...
+
+    private final User qwerty = new User( 3L, "qwerty", "qwerty", "qwerty", "qwerty@qwerty.com", "Qwerty!1", "qwerty");
+
+    @Test(expected = CredentialsExistException.class)
+    public void shouldThrowErrorWhenUsernameExists() {
+        //given user adds username
+        given(userRepository.findByUsername("qwerty")).willReturn(Optional.of(qwerty));
+        UserRequestDto copycat = createUserRequestDto("notqwerty","notqwerty", "notqwerty@qwerty.com", "notqwerty", "NotQwerty!1",  "qwerty");
+
+        //when username already exists
```
H-Ford: Your tests say what you've written in the comments in this test file.

H-Ford (commented):
Good work, just combine the Invalid format exceptions as we've discussed and get rid of the unnecessary comments

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -15,22 +18,62 @@
 
     private final UserRepository userRepository;
 
+    public static final String PASSWORD_PATTERN = "((?=.*[a-z])(?=.*[0-9])(?=.*[!?\\#@^&£$*+;:~])(?=.*[A-Z]).{8,16})";
+    public static final String EMAIL_PATTERN = "^[a-zA-Z0-9\\.\\!\\#\\$\\%\\&\\'\\*\\+\\-\\/\\=\\?\\^\\_\\`]+@[a-zA-Z0-9]+\\.[\\.A-Za-z]{1,10}";
+    public static final String USERNAME_PATTERN = "^[a-zA-Z0-9]*$";
```
ZaryabK: Would be nice to have comments on these regexs explaining what they're doing/how they're doing it - if the requirements in the future change for valid usernames/emails/passwords then this wouldn't be very maintainable!

Something that might help: https://alexwlchan.net/2016/04/regexes-are-code/

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -15,22 +18,62 @@
...
 
     @Override
     public UserResponseDto saveUser(UserRequestDto userRequestDto) {
-        User user = from(userRequestDto);
-        String password = user.getPassword();
+        String password = userRequestDto.getPassword();
+        String username = userRequestDto.getUsername();
+        String email = userRequestDto.getEmail();
+
+        usernameExists(username);
```
ZaryabK: Is there any reason you're checking username validity before email validity?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -15,22 +18,62 @@
...
+        emailExists(email);
+
         checkPassword(password);
-        user.setPassword(PasswordUtils.hash(password));
+        checkUsername(username);
+        checkEmail(email);
```
ZaryabK: Seems you're checking the email/username multiple times - these could be extracted out into their own methods (and both verification methods could be extracted out into it's own method, maybe something like `verifyUser`)?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -15,22 +18,62 @@
 
     private final UserRepository userRepository;
 
+    public static final String PASSWORD_PATTERN = "((?=.*[a-z])(?=.*[0-9])(?=.*[!?\\#@^&£$*+;:~])(?=.*[A-Z]).{8,16})";
```
ZaryabK: Question: should the regexs be in this class? Maybe somewhere like AuthUtils could be better (not so sure about this though)

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void shouldCheckPasswordDoesNotMatchHashedPassword() {
+        //given
+        String password = "password123";
+
+        //when
+        String hashedPassword = AuthUtils.hash(password);
+
+        //then
+        Assert.assertFalse(AuthUtils.verifyHash("WrongPassword", hashedPassword));
```
ZaryabK, line 79: I think this assertion is returning a false positive - you're using another method in the same class you're testing to test this method

Actually on further inspection it turns out all these methods are doing is using a method from another library - if you're not adding any extra logic then do these tests even have to be here? (open question)

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
ZaryabK: On a more general note, if you code is clear enough to read (as it should be!) then you don't need any comments at all!

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.gradproject2019.users.service.UserServiceImpl;
+import com.gradproject2019.utils.AuthUtils;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AuthUtilsTest {
+
+    @Test
+    public void shouldCheckPasswordIsValidAgainstRegex() {
```
ZaryabK, line 14: This is a testing antipattern - generally you shouldn't couple tests like this, if someone changed your regexs and one of these assertions were to fail, the whole test would fail and you'd have no idea which assertions were failing!

The solution to this would be to split out all of these tests into their own tests, with very descriptive names i.e. `shouldFailValidationIfPasswordHasNoSpecialCharacters" (applies to all fields)

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -15,4 +23,39 @@
...
+        userService.saveUser(copycat);
+    }
+
+    @Test(expected = CredentialsExistException.class)
+    public void shouldThrowErrorWhenEmailExists() {
+        //given user adds email
+        given(userRepository.findByEmail("qwerty@qwerty.com")).willReturn(Optional.of(qwerty));
+        UserRequestDto copycat = createUserRequestDto("notqwerty","notqwerty", "qwerty@qwerty.com", "notqwerty", "NotQwerty!1",  "notqwerty");
```
ZaryabK: I see these magic strings are being used in more than one place, so they should be extracted into their own `private static` fields

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -15,4 +23,39 @@
...
+
+        //when email already exists
+        userService.saveUser(copycat);
+    }
+
+    private UserRequestDto createUserRequestDto(String firstName, String lastName, String email, String occupation, String password, String username) {
```
ZaryabK: You could use this builder to create the copycat UserRequestDtos above!

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
ZaryabK, line 14: Is an empty password valid? (applies to all fields)

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
ZaryabK: Also for extra clarify something like `PASSWORD_VALIDATION_PATTERN" would be a better name, to make it clear what it's used for, and what it should be used for

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void shouldCheckUsernameNoSpaces() {
+        Assert.assertFalse(AuthUtils.validate("Sophia Grace", UserServiceImpl.USERNAME_PATTERN));
+        Assert.assertTrue(AuthUtils.validate("SophiaGrace", UserServiceImpl.USERNAME_PATTERN));
+    }
+
+    @Test
+    public void shouldCheckPasswordMatchesHashedPassword() {
+        //given
+        String password = "password123";
```
ZaryabK: This could be extracted into it's own final static field

ZaryabK (changes requested):

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
H-Ford: Does it matter?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
H-Ford: Comments aren't needed, tests explain the implementation

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
H-Ford: No because of the length assertion

H-Ford: The point of AssertJ is that it gives you more detailed messages about which assertion failed, the test class will be verbose with all the tests split out. Don't think this is needed.

karenzamani: Hi Zabi, I think we originally had our tests split out to test each particular regex requirement (i.e one for capital letters and another for special characters) but this resulted in a lot of duplicates. It probably because we weren't testing in the most effective way.
However I believe the above was constructed with guidance from Harry suggesting to remove individual tests and combine into one test. I understand what you're saying about testing each requirement in the regex, perhaps there is a better way to do this without duplication?

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
karenzamani: This has been amended

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
karenzamani: Done!

SpringProject/src/test/java/com/gradproject2019/AuthUtilsTest.java
karenzamani: Done!

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
karenzamani: Hi. The names have been changed for password, username and email regex to include the extra clarity as suggested. Originally we had the regexs in AuthUtils but decided they might be better placed in UserServiceImpl as that is where they are being used

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
karenzamani: Done!

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
karenzamani: Brief comments added into the code and the ticket updated in the board

SpringProject/src/main/java/com/gradproject2019/users/exception/InvalidRegexFormatException.java
```diff
@@ -0,0 +1,8 @@
+package com.gradproject2019.users.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(code = HttpStatus.BAD_REQUEST, reason="Invalid registration credentials.")
+public class InvalidRegexFormatException extends RuntimeException {
```
Piotrek605: The regex is something you define in your code, if it's invalid that means that you defined it wrong in your code. I think you meant InvalidCredentialsFormatException.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -15,22 +16,50 @@
...
+    //Username must not contain spaces
+    //Username can only contain lowercase, uppercase and numbers
+
     public UserServiceImpl(UserRepository userRepository) {
         this.userRepository = userRepository;
     }
 
     @Override
     public UserResponseDto saveUser(UserRequestDto userRequestDto) {
+        checkRegexValidity(userRequestDto);
```
Piotrek605: Same here, you don't check the regex's validity, you check the credentials' validity.

Piotrek605 (changes requested):

H-Ford (approved):

Piotrek605 (approved):

PR #19, opened 2019-11-25
Gp19 85 delete token endpoint

GBurley-Jones:
I am not sure if this is correct, as I think it needs to use a HTTP header from the front end to check if the correct token is being deleted. But any feedback would be welcome
Thanks!


Piotrek605 (approved):

Piotrek605 (approved):

singhals-els (approved):

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -94,21 +105,55 @@ public void shouldReturn401WhenUserExistsAndPasswordIncorrect() {
...
         Assert.assertEquals("Invalid user credentials.", response.getBody().getMessage());
     }
 
-    private ResponseEntity<Token> login(URI loginUri, HttpEntity<LoginDto> request) {
-        return restTemplate.exchange(loginUri, POST, request, Token.class);
+    //test to check the delete token endpoint
+    @Test
+    public void shouldReturn204andDeleteTokenWhenTokenExists() {
+        //given token exists
+        userRepository.saveAndFlush(user);
```
H-Ford: If you've already saved a user in your before method, you shouldn't need to save one here

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -39,33 +42,41 @@
...
     public void tearDown() {
         clearRepositories();
     }
 
+    //tests to check login in endpoint works
     @Test
     public void shouldReturn200andCreateTokenWhenUserExistsAndPasswordCorrect() {
         //given
-        HttpEntity<LoginDto> request = new HttpEntity<>(loginDto);
+        User savedUser = userRepository.saveAndFlush(user);
```
H-Ford: Same here with saved user

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -36,37 +44,48 @@
...
     private LoginDto loginDto;
+    private URI logoutUri;
+    private Token testToken;
+    private LoginDto secondLoginDto;
 
     @Before
     public void setUp() throws URISyntaxException {
         clearRepositories();
+
+        //hashedPassword = PasswordUtils.hash("P455w0rd!");
```
ZaryabK, line 60: You can remove this comment if it's not being used

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -36,37 +44,48 @@
...
+
+        testToken = new Token(savedUser.getId(), randomUUID());
     }
 
     @After
     public void tearDown() {
         clearRepositories();
     }
 
+    //tests to check login in endpoint works
```
ZaryabK, line 82: You can remove this comment as it doesn't really add to this test file (and similar comments above tests)

Ideally your test name would be descriptive enough that you wouldn't even need something like this!

SpringProject/src/it/java/com/gradproject2019/auth/AuthControllerIT.java
```diff
@@ -36,37 +44,48 @@
...
     @Test
     public void shouldReturn200andCreateTokenWhenUserExistsAndPasswordCorrect() {
         //given
```
ZaryabK, line 85: If your set up is all happening in the setUp() method, then you can remove the given comments in each test where no further steps are being taken

SpringProject/src/main/java/com/gradproject2019/auth/repository/AuthRepository.java
```diff
@@ -2,8 +2,18 @@
...
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.UUID;
 
 @Repository
-public interface AuthRepository extends JpaRepository<Token, Long> {
+public interface AuthRepository extends JpaRepository<Token, UUID> {
+//    @Modifying
```
ZaryabK, line 15: Is this supposed to be commented out? I would imagine this is _the_ method for this PR 😂

SpringProject/src/main/java/com/gradproject2019/auth/service/AuthServiceImpl.java
```diff
@@ -45,4 +46,16 @@ private Token createToken(Long userId) {
                 .withToken(UUID.randomUUID())
                 .build();
     }
+
+    @Override
+    public void logout(UUID token) {
+        checkTokenExists(token);
+        authRepository.deleteById(token);
```
ZaryabK, line 53: Is this method named correctly? It seems you're trying to delete by ID, but passing a token through instead

PR #21, opened 2019-11-26
GP19-98: verify authorization headers on conference endpoints + tests

MerrettM:


SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -181,11 +174,26 @@ public void shouldReturn204AndDeleteConference() throws URISyntaxException {
         Assert.assertFalse(conferenceRepository.existsById(savedConference.getId()));
     }
 
+    @Test
+    public void shouldReturn401WhenHeaderNotValid() throws URISyntaxException {
```
Piotrek605: Change to 'when token not valid' or 'unauthorized'

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -75,4 +85,12 @@ private void checkConferenceExists(Long conferenceId) {
             throw new ConferenceNotFoundException();
         }
     }
+
+    private void checkUserAuthorised(UUID token) {
+        try {
+            authServiceImpl.checkTokenExists(token);
+        } catch (TokenNotFoundException e) {
+            throw new InvalidCredentialsException();
```
Piotrek605: I think this one requires a new exception, something like UserUnauthorizedException. InvalidCredentialsException is already used for logging in with invalid credentials.

Piotrek605 (changes requested):

Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/utils/TestUtils.java
```diff
@@ -11,8 +18,29 @@
...
+
+    @Autowired
+    public TestRestTemplate restTemplate;
+
+    public User user;
+    public User savedUser;
+    public String hashedPassword;
+    public Token testToken;
+
     public void clearRepositories() {
```
H-Ford, line 35: Can be private

SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -198,10 +206,10 @@ public void shouldReturn404WhenConferenceToBeEditedNotFound() throws URISyntaxEx
     public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException {
         //given
         Conference savedConference = conferenceRepository.saveAndFlush(conference);
-        URI uri = new URI(baseUrl + "/" + savedConference.getId());
+        URI uri = new URI(baseUri + "/" + savedConference.getId());
         String newName= "Harry's conference";
         String newCity= "Manchester/North";
-        HttpEntity<ConferencePatchRequestDto> request = new HttpEntity<>(createPatchRequestDto(newName,null,newCity, null, null));
+        ConferencePatchRequestDto request = createPatchRequestDto(newName,null,newCity, null, null);
```
H-Ford: Spaces

MerrettM:
Just to clarify since it doesn't make sense at the moment: `commit 840e4202aa4856b496be187a644b9a3ffdeb489c` is incorrectly named, should say `"GP19-98: WIP edit user authorisation"`

SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -224,16 +194,28 @@ public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException
         return restTemplate.exchange(uri, GET, null, new ParameterizedTypeReference<ConferenceResponseDto>() {});
     }
 
-    private ResponseEntity<String> postConference(URI uri, HttpEntity<ConferenceRequestDto> request) {
-        return restTemplate.exchange(uri, POST, request, new ParameterizedTypeReference<String>() {});
+    private ResponseEntity<String> postConference(URI uri, ConferenceRequestDto request) {
+        HttpHeaders headers = new HttpHeaders();
+        headers.add(HttpHeaders.AUTHORIZATION, testToken.getToken().toString());
```
H-Ford: You could extract constructing the headers as a method

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -4,137 +4,109 @@
...
     @Test
     public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException {
-        //given
-        User savedUser = userRepository.saveAndFlush(user);
-        URI uri = new URI(baseUrl + "/" + savedUser.getId());
+        URI uri = new URI(baseUri + "/" + savedUser.getId());
         String newUsername= "sophiaUsername";
         String newFirstName= "Sophia";
-        HttpEntity<UserPatchRequestDto> request = new HttpEntity<>(createPatchRequestDto(newUsername,newFirstName,null, null, null, null));
+        UserPatchRequestDto request = createPatchRequestDto(newUsername, newFirstName,null, null, null);
```
H-Ford: Spaces

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -144,12 +116,31 @@ public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException
...
+
     private ResponseEntity<String> postUser(URI uri, HttpEntity<UserRequestDto> request) {
         return restTemplate.exchange(uri, POST, request, new ParameterizedTypeReference<String>() {});
     }
 
-    private ResponseEntity<UserResponseDto> editUser(URI uri, HttpEntity<UserPatchRequestDto> request) {
-        return restTemplate.exchange(uri, PATCH, request, new ParameterizedTypeReference<UserResponseDto>() {});
+    private ResponseEntity<UserResponseDto> editUser(URI uri, UserPatchRequestDto request) {
+        HttpHeaders headers = new HttpHeaders();
+        headers.add(HttpHeaders.AUTHORIZATION, testToken.getToken().toString());
```
H-Ford: Again extract

SpringProject/src/main/java/com/gradproject2019/conferences/repository/ConferenceRepository.java
```diff
@@ -14,8 +14,14 @@
...
-    @Query("UPDATE Conference c SET c.name = IfNull(:name, c.name), c.dateTime = IfNull(:dateTime, c.dateTime), c.city = IfNull(:city, c.city), c.description = IfNull(:description, c.description), c.topic = IfNull(:topic, c.topic) WHERE c.id = :id")
+    @Modifying(clearAutomatically = true)
+    @Query("UPDATE Conference c SET " +
+            "c.name = IfNull(:name, c.name), " +
+            "c.dateTime = IfNull(:dateTime, c.dateTime), " +
+            "c.city = IfNull(:city, c.city), " +
+            "c.description = IfNull(:description, c.description), " +
+            "c.topic = IfNull(:topic, c.topic) " +
+            "WHERE c.id = :id")
     void updateConference(@Param("id") Long id, @Param("name") String name, @Param("dateTime") Instant dateTime, @Param("city") String city, @Param("description") String description, @Param("topic") String topic);
```
H-Ford, line 31: Put some of these parameters on a new line

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -49,26 +85,42 @@ private void checkCredentials(String input, String pattern) {
...
 
-        userRepository.updateUser(userId, userPatch.getFirstName(), userPatch.getLastName(), userPatch.getUsername(), userPatch.getEmail(), userPatch.getOccupation());
+    private Token getTokenByUUID(UUID token) {
+        return authRepository
+                .findById(token)
+                .orElseThrow(UserUnauthorisedException::new);
+    }
 
-        return getUserById(userId);
+    private void checkTokenMatchesUser(Long tokenUserId, Long userId) {
```
Piotrek605: The method is called 'check token matches user' but you're actually passing two user ids to it. It would be cleaner if you passed a token and a user id or if you named it 'check if user ids match'.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -1,46 +1,82 @@
...
+
+        userRepository.updateUser(userId, userPatch.getFirstName(), userPatch.getLastName(), patchUsername, patchEmail, userPatch.getOccupation());
+        return new UserResponseDto().from(getUserById(userId));
+    }
+
+    private void checkPasswordValidity(String password) {
+        checkCredentials(password, PASSWORD_VALIDATION_PATTERN);
+    }
+
+    private void checkUsernameAndEmailValidity(String username, String email) {
```
Piotrek605: Is this method needed? Why not just call checkCredentials and do the null check there?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -49,26 +85,42 @@ private void checkCredentials(String input, String pattern) {
...
-    public UserResponseDto getUserById(Long userId) {
-        return new UserResponseDto().from(userRepository
-                .findById(userId)
-                .orElseThrow(UserNotFoundException::new));
+    private void checkIfUpdatedCredentialsExist(String patchUsername, String patchEmail, User user) {
+        if(patchUsername != null && !patchUsername.equals(user.getUsername())) {
+            checkIfUsernameExists(patchUsername);
+        }
+        if(patchEmail != null && !patchEmail.equals(user.getEmail())) {
+            checkIfUsernameExists(patchEmail);
```
Piotrek605: This should be checkIfEmailExists, shouldn't it?

Piotrek605 (changes requested):

Piotrek605:
IT tests need some refactoring, check the Slack discussion for this. You can just change the names of the tests added in this PR and create a ticket for further refactoring if you wish.

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -144,12 +116,27 @@ public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException
         Assert.assertEquals(savedUser.getOccupation(), response.getBody().getOccupation());
     }
 
+    @Test
+    public void shouldReturn401WhenUserUnauthorised() throws URISyntaxException {
```
Piotrek605: Rename to something like shouldReturn401WhenUnauthorizedEdit

SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -4,209 +4,179 @@
...
         ResponseEntity<String> response = deleteConference(uri);
 
-        //then
-        Assert.assertEquals(204,response.getStatusCodeValue());
+        Assert.assertEquals(204, response.getStatusCodeValue());
         Assert.assertFalse(conferenceRepository.existsById(savedConference.getId()));
     }
 
+    @Test
+    public void shouldReturn401WhenUnauthorised() throws URISyntaxException {
```
Piotrek605: Rename to something like shouldReturn401WhenUnauthorizedDelete to avoid a clash with shouldReturn401WhenUnauthorizedEdit.

Piotrek605 (changes requested):

Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -1,131 +1,97 @@
...
-        userRepository.deleteAll();
+        clearRepositories();
     }
 
     @Test
     public void shouldReturn200AndSaveUserInDatabase() throws URISyntaxException {
-        URI uri = new URI(baseUrl);
-        UserRequestDto userRequestDto = createRequestDto("GracesCoolUsername", "Grace", "Burley Jones", "Grace@gmail.com", "P455w0rd!", "Botanist");
+        URI uri = new URI(baseUri);
+        UserRequestDto userRequestDto = createRequestDto("MichaelsCoolUsername", "Michael", "Merrett", "MGHMerrett@gmail.com", "P455w0rd!", "Frustrated");
```
ed-steele-rdm-els, line 49: :eyes:

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -14,9 +14,20 @@
 public interface UserRepository extends JpaRepository<User, Long> {
 
     @Transactional
-    @Modifying
-    @Query("UPDATE User u SET u.firstName = IfNull(:firstName, u.firstName), u.lastName = IfNull(:lastName, u.lastName), u.username = IfNull(:username, u.username), u.email = IfNull(:email, u.email), u.occupation = IfNull(:occupation, u.occupation) WHERE u.id = :id")
-    void updateUser(@Param("id") Long id, @Param("firstName") String firstName, @Param("lastName") String lastName, @Param("username") String username, @Param("email") String email, @Param("occupation") String occupation);
+    @Modifying(clearAutomatically = true)
+    @Query("UPDATE User u SET " +
+            "u.firstName = IfNull(:firstName, u.firstName), " +
```
ed-steele-rdm-els: you don't need table aliasing here (or in the other one). It doesn't hurt to have it but it doesn't look right

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -75,4 +85,12 @@ private void checkConferenceExists(Long conferenceId) {
             throw new ConferenceNotFoundException();
         }
     }
+
```
ed-steele-rdm-els, line 88: (sorry if double comment) 

I suggest having this check at the web/resource layer. 

Having it here means that an authentication token needs to be provided to do these business functions. If another part of the application wants to edit/save/delete a conference using these APIs they will need to have to provide a token, despite them being invoked presumably after authentication

The Business layer shouldn't really handle `authentication` related things, it should only be aware of business related functions - it really helps to keep each 'tier' of the app separate

MerrettM: So I know that the business layer translates to the Service layer since it does the business logic, but what layer does the web/resource layer translate to? from the "web" part of the name I'm guessing it's the front end (which would then require each action that requires authentication to hit to both an authorisation endpoint and the endpoint that carries out said action). But if it's the back end then the authentication would have to happen in the Controller layer which we've been told to keep logic out of?

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
MerrettM: I'm assuming you are referring to the `User u` and `u.firstName` bits?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
MerrettM: It's been a long day... 😄

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
ed-steele-rdm-els: correcto

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
ed-steele-rdm-els: In this case it refers to your `*Controllers` - its the layer that takes HTTP requests and does something with them. It is the API (_interface_) that your application has to the wider internet (FE, or other 3rd party applications). You _cannot_ rely on a front end to provide security

Keeping logic out the `Controller` is an important consideration - if its what you've been told by your helpers you should follow their advice. 

However when you think about it, your controllers already have logic in - 
`@RequestBody UserPatchRequestDto userPatchRequestDto, UserRequestDto userRequestDto, @PathVariable("id") Long userId`

This Deserializes a HTTP request into java objects (and does type checking) - you don't pass a stream of bytes say into the business layer. I would suggest creating an `abstract class` called `BaseController` or something that can hande authentication across _all_ controllers with a common `checkAuthentication(UUID token)` method

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -44,44 +45,72 @@ public UserResponseDto saveUser(UserRequestDto userRequestDto) {
...
+        userRepository.updateUser(userId, userPatchRequestDto.getFirstName(), userPatchRequestDto.getLastName(), userPatchRequestDto.getUsername(), userPatchRequestDto.getEmail(), userPatchRequestDto.getOccupation());
+        return new UserResponseDto().from(getUserById(userId));
     }
 
-    private void checkCredentialsValidity(UserRequestDto userRequestDto) {
+    private void checkValidSave(UserRequestDto userRequestDto) {
         checkCredentials(userRequestDto.getPassword(), PASSWORD_VALIDATION_PATTERN);
         checkCredentials(userRequestDto.getUsername(), USERNAME_VALIDATION_PATTERN);
         checkCredentials(userRequestDto.getEmail(), EMAIL_VALIDATION_PATTERN);
+        checkIfUsernameExists(userRequestDto.getUsername());
```
ed-steele-rdm-els, line 64: Look into using an `UNIQUE` index here 
https://dev.mysql.com/doc/refman/8.0/en/constraint-primary-key.html
You'll need to have another migration script that adds a `UNQIUE` index to the username and email columns. https://dev.mysql.com/doc/refman/8.0/en/alter-table.html (ask for more details if the syntax isn't clear, or use MySqlWorkbench to do it graphically)

It stops multiple emails/usernames being created at the database level.

(You may not understand this yet) 
The issue with handling duplicates at the application level is that this app with have many threads communicating with a single db instance. If 2 threads are working at the same time (and with the same email/username), you could have a situation with 2 identical usernames in the db. 

When they both query the db (_before_) writing, this check will return true - there are no usernames like that. They could then _both_ write to the db at exactly, or almost the same time, which would be succesful in both cases (without `UNIQUE` indexes) 

After this any time the user/users tried to use your app they wouldn't be able to do anything because any time they did - `checkIfUsernameExists(String username)` would return false

H-Ford (approved):

PR #22, opened 2019-11-27
Gp19 88 user by Id

GBurley-Jones:


SpringProject/src/main/java/com/gradproject2019/auth/service/AuthServiceImpl.java
```diff
@@ -53,9 +53,9 @@ public void logout(UUID token) {
         authRepository.deleteById(token);
     }
 
-    private void checkTokenExists(UUID token) {
+    public boolean checkTokenExists(UUID token) {
         if(!authRepository.findById(token).isPresent()) {
             throw new TokenNotFoundException();
-        }
+        } return(true);
```
Piotrek605: Some whitespace issues here, this should be
}
return true;

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -35,6 +43,14 @@ public UserResponseDto saveUser(UserRequestDto userRequestDto) {
         return new UserResponseDto().from(userRepository.saveAndFlush(user));
     }
 
+    @Override
+    public UserResponseDto findUserById(Long userId, UUID token) {
+        userExistsById(userId);
```
Piotrek605: I don't think this is necessary, getUserById will also throw an exception if the user doesn't exist.

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        Assert.assertEquals(savedUser.getId(), response.getBody().getId());
+        Assert.assertEquals(newUsername, response.getBody().getUsername());
+        Assert.assertEquals(newFirstName, response.getBody().getFirstName());
+        Assert.assertEquals(savedUser.getLastName(), response.getBody().getLastName());
+        Assert.assertEquals(savedUser.getEmail(), response.getBody().getEmail());
+        Assert.assertEquals(savedUser.getOccupation(), response.getBody().getOccupation());
+    }
+
+    @Test
+    public void shouldReturn200WhenUserRegisteredAndLoggedIn() throws URISyntaxException {
```
Piotrek605: The name doesn't say what the test does, it should be something like 'should return 200 when getting user by id'

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        Token testToken = new Token(savedUser.getId(), randomUUID());
+        Token savedToken = authRepository.saveAndFlush(testToken);
+
+        ResponseEntity<User> response = getUserById(savedUser.getId(), savedToken.getToken());
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+    }
+
+    @Test
+    public void shouldReturn404WhenUserNotLoggedIn() throws URISyntaxException{
```
Piotrek605: Same here, this should be something like 'should return 404 when getting user by id unauthorized', you don't need to specify what unauthorized means

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -24,4 +24,16 @@
         UserResponseDto newUser = userService.saveUser(userRequestDto);
         return ResponseEntity.ok(newUser);
     }
+
+    @PatchMapping(path = "/users/{id}")
+    public ResponseEntity<UserResponseDto> editUser(@RequestBody UserPatchRequestDto userPatchRequestDto, UserRequestDto userRequestDto, @PathVariable("id") Long userId) {
```
Piotrek605: Why do you need both UserPatchRequestDto and UserRequestDto? What about the token for authorization?

Piotrek605 (changes requested):

.idea/grads19-be.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vsocolov-elsevier, line 1: This file should not be pushed. It should be ignored in git

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vsocolov-elsevier, line 1: This file as well should be ignored.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vsocolov-elsevier, line 1: Same here. Everything should be ignored in .idea directory

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
GBurley-Jones: This is something outside the scope of this ticket. @MerrettM is working on this in a separate ticket

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
GBurley-Jones: Thank you, I thought I'd fixed this in the merge!

.idea/grads19-be.iml
GBurley-Jones: I have added this to the .gitignore file

Screenshot 2019-11-27 at 12 42 26

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
Piotrek605: That answers the second part of the question, what about the first one?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -35,6 +43,14 @@ public UserResponseDto saveUser(UserRequestDto userRequestDto) {
         return new UserResponseDto().from(userRepository.saveAndFlush(user));
     }
 
+    @Override
+    public UserResponseDto findUserById(Long userId, UUID token) {
+        getUserById(userId);
```
Piotrek605: Again, this call is unnecessary.

Piotrek605 (changes requested):

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
MerrettM: Both of those are part of my ticket's scope, the first part of the question is because that's wrong and I've fixed it, the second part is that I'm currently implementing this in my PR.

.idea/grads19-be.iml
ZaryabK: You may need to push your updated gitignore, or remove these from the PR

Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        authRepository.deleteAll();
         userRepository.deleteAll();
     }
 
     @Test
     public void shouldReturn200AndSaveUserInDatabase() throws URISyntaxException {
-        //given
         URI uri = new URI(baseUrl);
-        UserRequestDto userRequestDto = createRequestDto("KaramsCoolUsername", "Karam", "Kapoor", "KSinghK@gmail.com", "P455w0rd!", "Botanist");
+        UserRequestDto userRequestDto = createRequestDto("GracesCoolUsername", "Grace", "Burley Jones", "Grace@gmail.com", "P455w0rd!", "Botanist");
```
ZaryabK, line 72: Is there a reason this user had to change?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
-        //when
         ResponseEntity<String> response = postUser(uri, request);
 
-        //Then
         Assert.assertEquals(400,response.getStatusCodeValue());
     }
 
+    @Test
+    public void shouldReturn404WhenUserToBeEditedNotFound() throws URISyntaxException{
+        URI uri = new URI(baseUrl + "/1000000000");
```
ZaryabK: Just a suggestion: in the real world you would never make a URI like this because you're making an assumption that you won't have 1000000000 users

If you want to make sure it's a user that'll never be found you could do one of 2 things:
1. baseUrl + "/" + Integer.MAX_VALUE 
2. use a number that you know will never be used like 0 or -1 (assuming your API handles these things gracefully)

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        HttpEntity<UserPatchRequestDto> request = new HttpEntity<>(createPatchRequestDto(null, null, null, null, null, null));
```
ZaryabK, line 115: Do you have a patch request DTO builder? Would be useful for building the patch requests in these tests

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        Assert.assertEquals(newUsername, response.getBody().getUsername());
+        Assert.assertEquals(newFirstName, response.getBody().getFirstName());
+        Assert.assertEquals(savedUser.getLastName(), response.getBody().getLastName());
+        Assert.assertEquals(savedUser.getEmail(), response.getBody().getEmail());
+        Assert.assertEquals(savedUser.getOccupation(), response.getBody().getOccupation());
+    }
+
+    @Test
+    public void shouldReturn200WhenGettingUserById() throws URISyntaxException {
+        User user = new User( "Grace", "Grace", "Jones", "gbj@gmail.com", "P455w0rd!", "na");
```
ZaryabK: Again hoping this isn't a real email address!

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+        Token savedToken = authRepository.saveAndFlush(testToken);
+
+        ResponseEntity<User> response = getUserById(savedUser.getId(), savedToken.getToken());
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+    }
+
+    @Test
+    public void shouldReturn404WhenGettingUserByIDUnauthorised() throws URISyntaxException{
+        User user = new User( "Grace", "Grace", "Jones", "gbj@gmail.com", "P455w0rd!", "na");
```
ZaryabK: You're creating the same user here as above, maybe this could be extracted into the set up method?

.idea/grads19-be.iml
GBurley-Jones: This has now been fixed on @MerrettM's branch. We will double check these files are not present once they have both the branches have been merged

SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -59,154 +59,120 @@ public void tearDown() {
...
-        //given
         URI uri = new URI(baseUrl);
         Conference addedConference = conferenceRepository.saveAndFlush(conference);
 
-        //when
         ResponseEntity<List<ConferenceResponseDto>> response = getConferenceList(uri);
 
-        //Then
         Assert.assertEquals(200,response.getStatusCodeValue());
         Assert.assertEquals(conference.getName(),response.getBody().get(0).getName());
```
H-Ford: Instead of doing .get(0) you can use extracting() and extract the fields you want to assert on from the ConferenceResponseDto

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -30,87 +37,146 @@
...
+    public void shouldReturn200WhenGettingUserById() throws URISyntaxException {
+        User user = new User( "Grace", "Grace", "Jones", "gbj@gmail.com", "P455w0rd!", "na");
+        User savedUser = userRepository.saveAndFlush(user);
+
+        Token testToken = new Token(savedUser.getId(), randomUUID());
+        Token savedToken = authRepository.saveAndFlush(testToken);
+
+        ResponseEntity<User> response = getUserById(savedUser.getId(), savedToken.getToken());
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
```
H-Ford, line 150: You should assert on at least some of the fields here, not just the response code

SpringProject/src/main/java/com/gradproject2019/auth/service/AuthServiceImpl.java
```diff
@@ -53,9 +53,10 @@ public void logout(UUID token) {
         authRepository.deleteById(token);
     }
 
-    private void checkTokenExists(UUID token) {
+    public boolean checkTokenExists(UUID token) {
```
H-Ford: the naming convention for methods returning boolean is for them to start with is. For example isExistingToken

H-Ford: Does this need to return a boolean though

H-Ford (changes requested):
Few things to change otherwise great work!

karenzamani (approved):

ZaryabK (changes requested):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
GBurley-Jones: This was just a test to make sure that the user being checked wasn't the same as the one already in the database. But this is not really needed anymore

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
GBurley-Jones: There already is a builder for the patch, this is a private method using the builder

PR #23, opened 2019-12-02
GP19-121: added GET endpoint to list all users

KaramKapoor:


SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -19,6 +21,12 @@
     @Autowired
     private UserService userService;
 
+    @GetMapping(path = "/users")
```
vsocolov-elsevier: I think it's a good practice to provide an optional params for paginating this list (page, pageSize). If the number of users will increase dramatically, this endpoint will become unusable.

Piotrek605 (approved):

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
KaramKapoor: At the moment the users will be listed in a scrollable dropdown box, and in the future it should be replaced by a search box. Would this mean that pagination is not required yet?

vsocolov-elsevier: In that case pagination is mandatory. Imagine that you have in your system 10K users, if so, all 10K users will be in this scroll box which will freeze entire web page.

KaramKapoor: Ok thanks i'll add in pagination, but still a bit confused about its implementation in the dropdown box - will the scrolling be replaced by pages of users?

vsocolov-elsevier: Yep, in the web you'll request batch of users to show in dropdown. First just 10-20 users, after that while scrolling another request for next 10 users and so on.

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -141,6 +144,28 @@ public void shouldReturn401WhenUnauthorisedEdit() throws URISyntaxException {
         Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
     }
 
+    @Test
+    public void shouldReturn200AndEmptyListWhenDatabaseEmpty() throws URISyntaxException {
+        URI uri = new URI(baseUri);
```
ZaryabK: You're creating this URI in multiple tests, so it can be extracted

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -141,6 +144,28 @@ public void shouldReturn401WhenUnauthorisedEdit() throws URISyntaxException {
...
+        clearRepositories();
```
ZaryabK, line 140: Can't see it in this PR but this is happening in the 'before' step for every test so this is unnecessary(?), unless I'm missing something haha

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -141,6 +144,28 @@ public void shouldReturn401WhenUnauthorisedEdit() throws URISyntaxException {
         Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
     }
 
+    @Test
+    public void shouldReturn200AndEmptyListWhenDatabaseEmpty() throws URISyntaxException {
```
ZaryabK: 'WhenDatabaseEmpty' is not very BDD friendly - maybe something like `WhenNoUsers` would be better

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -141,6 +144,28 @@ public void shouldReturn401WhenUnauthorisedEdit() throws URISyntaxException {
...
+        clearRepositories();
+
+        ResponseEntity<List<UserResponseDto>> response = getUserList(uri);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(true, response.getBody().isEmpty());
+    }
+
+    @Test
+    public void shouldReturn200AndListOfUsersWhenDatabasePopulated() throws URISyntaxException {
```
ZaryabK: Again the test name here could be changed to `WhenUserExists`

Seeing as you're only adding 1 user, it would be nice to assert the size is 1 here as well

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
ZaryabK: Just a note that it might be complex (but necessary) to test the pagination on the endpoint, so I would recommend speaking to your BE tech lead on the scope of the tests.

I would expect there to be at least 1 test to test that you can navigate the pages (forwards and backwards) and that the page returns the user(s) you'd expect.

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
KaramKapoor: This is for a test where we don't want anything in the database, so anything added to the database in the universalSetup() should be cleared before this test

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -29,10 +30,12 @@ void updateUser(@Param("id") Long id,
...
 
-    @Query(value = "SELECT * FROM users where username = :username", nativeQuery = true)
+    @Query(value = "SELECT * FROM users WHERE username = :username", nativeQuery = true)
     Optional<User> findByUsername(@Param("username") String username);
 
-    @Query(value = "SELECT * FROM users where email = :email", nativeQuery = true)
+    @Query(value = "SELECT * FROM users WHERE email = :email", nativeQuery = true)
     Optional<User> findByEmail(@Param("email") String email);
 
+    @Query(value = "SELECT * FROM users WHERE first_name like CONCAT('%', :query,'%') OR last_name like CONCAT('%', :query, '%') ORDER BY last_name ASC", nativeQuery = true)
```
ed-steele-rdm-els: nitpick: LIKE should be capitalized as is a MySql keyword
Can you format your queries like this
```
SELECT x,q,c 
FROM tablename
WHERE <>
AND <>
OR <>
ORDER BY foo
```
Its the standard way of formatting SQL

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -19,6 +20,12 @@
     @Autowired
     private UserService userService;
 
+    @GetMapping(path = "/users/search")
```
ed-steele-rdm-els, line 23: I'm not saying _don't_ do this - but if you look at https://internal-docs.staging.corp.mendeley.com/apidocs/docs (appreciate some things in their aren't perfect) see how these implement 'searching' e.g. profiles-service

Most have the base `/users/` path with optional query params to allow _filtering_ which strictly speaking is what you are doing. This is because `list` endpoints can and should be filterable/sortable on many columns - but this will make your code for the list dao more complex. 

For now, I'm a believer in making things work first, so I've got no problem with this in a sandbox project - but be aware its not the most ideal thing to do

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
ed-steele-rdm-els: also isn't what your doing more of a startswith? So the LIKE should be first_name 
`LIKE :query%`

KaramKapoor: Have formatted it as suggested. For the 'startswith', i think the current way is more user-friendly since a user may misspell or not remember part of their manager's name, but still get the results they want

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -19,6 +24,12 @@
     @Autowired
     private UserService userService;
 
+    @GetMapping(path = "/users/search/{searchInput}")
```
Piotrek605: Search input should be a query parameter, look at 
google.com/search?q=something
That's because the 'something' isn't pointing to a specific resource, it's just filtering the user resource

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
KaramKapoor: I agree with your point regarding the filtering, and I think it's something we might implement after the feature freeze when we're addressing our technical debt

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -30,12 +30,23 @@ void updateUser(@Param("id") Long id,
...
+    @Query(value = "SELECT * " +
+            "FROM users " +
+            "WHERE email = :email",
+            nativeQuery = true)
     Optional<User> findByEmail(@Param("email") String email);
 
-    @Query(value = "SELECT * FROM users WHERE first_name like CONCAT('%', :query,'%') OR last_name like CONCAT('%', :query, '%') ORDER BY last_name ASC", nativeQuery = true)
+    @Query(value = "SELECT * " +
+            "FROM users " +
+            "WHERE CONCAT(first_name, ' ', last_name)" +
```
ed-steele-rdm-els: make sure to leave spaces at the start of each line - this will all form one long string

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -33,6 +35,13 @@ public UserServiceImpl(UserRepository userRepository, AuthServiceImpl authServic
         this.authServiceImpl = authServiceImpl;
     }
 
+    @Override
+    public List<UserResponseDto> searchByName(String searchInput) {
+        return userRepository.searchByName(searchInput).stream()
+                .map(user -> UserResponseDto.UserResponseDtoBuilder.anUserResponseDto().withFirstName(user.getFirstName()).withLastName(user.getLastName()).withId(user.getId()).build())
```
Piotrek605, line 41: It's 'a user' not 'an user'

Piotrek605 (changes requested):

Piotrek605 (approved):

ZaryabK (approved):

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
ed-steele-rdm-els: sure - i didn't mean do it now - more just something to think about

PR #24, opened 2019-12-04
GP19-118: store manager ids

MerrettM:


SpringProject/src/main/resources/db/migration/V2__create_schema.sql
```diff
@@ -5,5 +5,6 @@ CREATE TABLE users (
     last_name VARCHAR(255),
     email VARCHAR(255),
     password CHAR(60),
-    occupation VARCHAR(255)
+    occupation VARCHAR(255),
```
ed-steele-rdm-els, line 8: As you are wiping your database every time you'll get away with this - but this is not how flyway works - you're meant to have another migration script that updates your table to add these new columns (meant for auditing purposes I think) - be careful

Piotrek605 (approved):

ed-steele-rdm-els (approved):
That was a lot of code changes for adding a single new column right? You've obviously done nothing wrong, its how the libraries you are using are _meant_ to work

PR #25, opened 2019-12-05
GP19-120 restrict ability to get user data

MerrettM:
Only a User, their manager, and any user they manage should be able to see their profile.

### DIFF
--- a/SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
@@ -1,5 +1,6 @@
 package com.gradproject2019.user;
 
+import com.gradproject2019.auth.persistence.Token;
 import com.gradproject2019.users.data.UserPatchRequestDto;
 import com.gradproject2019.users.data.UserRequestDto;
 import com.gradproject2019.users.data.UserResponseDto;
@@ -10,17 +11,20 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.jupiter.params.shadow.com.univocity.parsers.annotations.Headers;
 import org.junit.runner.RunWith;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.ResponseEntity;
 import org.springframework.test.context.junit4.SpringRunner;
 
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
+import java.util.UUID;
 
 import static org.springframework.http.HttpMethod.*;
 
@@ -126,8 +130,8 @@ public class UserControllerIT extends TestUtils {
     }
 
     @Test
-    public void shouldReturn200WhenGettingUserById() {
-        ResponseEntity<User> response = getUserById();
+    public void shouldReturn200WhenGettingUserByOwnId() {
+        ResponseEntity<User> response = getUserById(uri2, passingHeaders);
 
         Assert.assertEquals(200, response.getStatusCodeValue());
         Assert.assertEquals(savedUser.getId(), response.getBody().getId());
@@ -139,12 +143,58 @@ public class UserControllerIT extends TestUtils {
     }
 
     @Test
-    public void shouldReturn401WhenGettingUserByIdUnauthorised() {
+    public void shouldReturn200WhenGettingManagerById() {
+        User employee = userRepository.saveAndFlush(new User("employee", "employee", "employee", "employee@email.com", "Password!1", "employee", savedUser.getId()));
+        Token employeeToken = authRepository.saveAndFlush(new Token(employee.getId(), UUID.randomUUID()));
+        HttpHeaders employeeHeaders = new HttpHeaders();
+        employeeHeaders.add(HttpHeaders.AUTHORIZATION, employeeToken.getToken().toString());
+
+        ResponseEntity<User> response = getUserById(uri2, employeeHeaders);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(savedUser.getId(), response.getBody().getId());
+        Assert.assertEquals(savedUser.getFirstName(), response.getBody().getFirstName());
+        Assert.assertEquals(savedUser.getUsername(), response.getBody().getUsername());
+        Assert.assertEquals(savedUser.getLastName(), response.getBody().getLastName());
+        Assert.assertEquals(savedUser.getEmail(), response.getBody().getEmail());
+        Assert.assertEquals(savedUser.getOccupation(), response.getBody().getOccupation());
+    }
+
+    @Test
+    public void shouldReturn200WhenGettingEmployeeById() throws URISyntaxException {
+        User employee = userRepository.saveAndFlush(new User("employee", "employee", "employee", "employee@email.com", "Password!1", "employee", savedUser.getId()));
+        URI employeeUri = new URI(baseUri + "/" + employee.getId());
+
+        ResponseEntity<User> response = getUserById(employeeUri, passingHeaders);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(employee.getId(), response.getBody().getId());
+        Assert.assertEquals(employee.getFirstName(), response.getBody().getFirstName());
+        Assert.assertEquals(employee.getUsername(), response.getBody().getUsername());
+        Assert.assertEquals(employee.getLastName(), response.getBody().getLastName());
+        Assert.assertEquals(employee.getEmail(), response.getBody().getEmail());
+        Assert.assertEquals(employee.getOccupation(), response.getBody().getOccupation());
+    }
+
+    @Test
+    public void shouldReturn401WhenGettingUserByIdUnauthorisedDueToToken() {
         ResponseEntity<ErrorEntity> response = getUserByIdExpectingAuthError();
 
         Assert.assertEquals(401, response.getStatusCodeValue());
         Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
     }
+
+    @Test
+    public void shouldReturn403WhenGettingUserWhoIsNotTheUserOrTheirManagerOrTheirEmployee() throws URISyntaxException {
+        User unconnectedUser = userRepository.saveAndFlush(new User("unconnected", "unconnected", "unconnected", "unconnected@email.com", "Password!1", "unconnected", null));
+        URI unconnectedUri = new URI(baseUri + "/" + unconnectedUser.getId());
+
+        ResponseEntity<ErrorEntity> response = getUserByIdExpectingError(unconnectedUri);
+
+        Assert.assertEquals(403, response.getStatusCodeValue());
+        Assert.assertEquals("User forbidden to perform action.", response.getBody().getMessage());
+    }
+
 
     @Test
     public void shouldReturn401WhenUnauthorisedEdit() {
@@ -242,8 +292,12 @@ public class UserControllerIT extends TestUtils {
         return restTemplate.exchange(uri2, PATCH, new HttpEntity<>(request, failingHeaders), ErrorEntity.class);
     }
 
-    private ResponseEntity<User> getUserById() {
-        return restTemplate.exchange(uri2, GET, new HttpEntity<>(passingHeaders), User.class);
+    private ResponseEntity<User> getUserById(URI uri, HttpHeaders headers) {
+        return restTemplate.exchange(uri, GET, new HttpEntity<>(headers), User.class);
+    }
+
+    private ResponseEntity<ErrorEntity> getUserByIdExpectingError(URI uri) {
+        return restTemplate.exchange(uri, GET, new HttpEntity<>(passingHeaders), ErrorEntity.class);
     }
 
     private ResponseEntity<ErrorEntity> getUserByIdExpectingAuthError() {
--- a/SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
@@ -139,7 +139,7 @@ public class UserControllerIT extends TestUtils {
     }
 
     @Test
-    public void shouldReturn401WhenGettingUserByIDUnauthorised() {
+    public void shouldReturn401WhenGettingUserByIdUnauthorised() {
         ResponseEntity<ErrorEntity> response = getUserByIdExpectingAuthError();
 
         Assert.assertEquals(401, response.getStatusCodeValue());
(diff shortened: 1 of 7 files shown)


SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -122,6 +123,36 @@ private void checkTokenMatchesUser(UUID token, Long userId) {
...
+            } catch (UserUnauthorisedException | NullPointerException b) {
+                try {
+                checkManagerIdMatchesId(requestingUser, requestedUser);
+                } catch (NullPointerException c) {
+                    throw new UserUnauthorisedException();
+                }
+            }
+        }
+    }
+
```
MerrettM: I realise this is disgusting, I have a solution involving if statements and negating the need to catch null pointers which I thought of as I was leaving yesterday. This will be implemented on Monday next week.

MerrettM: This has been fixed in the latest commit.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -117,7 +118,39 @@ private void checkIfManagerExists(Long managerId) {
...
     private void checkTokenMatchesUser(UUID token, Long userId) {
-        if(!authServiceImpl.getTokenById(token).getUserId().equals(userId)) {
+        if(!tokenMatchesUser(token, userId)) {
+            throw new UserUnauthorisedException();
+        }
+    }
+
+    private boolean tokenMatchesUser(UUID token, Long userId) {
+        boolean tokenMatchesUser = false;
+        if(authServiceImpl.getTokenById(token).getUserId().equals(userId)) {
```
H-Ford: You can just return your check, no need for the tokenMatchesUser variable

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -117,7 +118,39 @@ private void checkIfManagerExists(Long managerId) {
...
+            tokenMatchesUser = true;
+        }
+        return tokenMatchesUser;
+    }
+
+    private boolean userIdMatchesManagerId(User requestingUser, User requestedUser) {
+        boolean userIdMatchesManagerId = false;
+        if(requestedUser.getManagerId() != null && requestingUser.getId().equals(requestedUser.getManagerId())) {
```
H-Ford: Likewise here no need for the extra variable

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -117,7 +118,39 @@ private void checkIfManagerExists(Long managerId) {
...
+            userIdMatchesManagerId = true;
+        }
+        return userIdMatchesManagerId;
+    }
+
+    private boolean managerIdMatchesUserId(User requestingUser, User requestedUser) {
+        boolean managerIdMatchesUserId = false;
+        if(requestingUser.getManagerId() != null && requestingUser.getManagerId().equals(requestedUser.getId())) {
```
H-Ford: Same here

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -117,8 +114,28 @@ private void checkIfManagerExists(Long managerId) {
     }
 
     private void checkTokenMatchesUser(UUID token, Long userId) {
-        if(!authServiceImpl.getTokenById(token).getUserId().equals(userId)) {
-            throw new UserUnauthorisedException();
+        if(!tokenMatchesUser(token, userId)) {
+            throw new UserRecognisedButUnauthorizedException();
```
Piotrek605: UserForbiddenException

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -117,8 +114,28 @@ private void checkIfManagerExists(Long managerId) {
...
+    }
+
+    private boolean managerIdMatchesUserId(User requestingUser, User requestedUser) {
+        return requestingUser.getManagerId() != null && requestingUser.getManagerId().equals(requestedUser.getId());
+    }
+
+    private void checkUserRequestingIsAuthorized(Long requestedUserId, UUID token) {
+        User requestingUser = getUserById(authServiceImpl.getTokenById(token).getUserId());
+        User requestedUser = getUserById(requestedUserId);
+        if (!tokenMatchesUser(token, requestedUserId) && !userIdMatchesManagerId(requestingUser, requestedUser) && !managerIdMatchesUserId(requestingUser, requestedUser)) {
```
Piotrek605: You can call checkTokenMatches user here instead of tokenMachesUser and then you won't need tokenMatchesUser as a seperate method?

Is this the only place where userIdMatchesManagerId and managerIdMatchesUserId are called? If it is, then you can put the code here and delete those methods because their names are a bit confusing.

Piotrek605 (changes requested):

Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -139,9 +139,7 @@ public void shouldReturn200WhenGettingUserById() {
     }
 
     @Test
-    public void shouldReturn401WhenGettingUserByIDUnauthorised() throws URISyntaxException {
-        URI uri = new URI(baseUri + "/" + savedUser.getId());
-
+    public void shouldReturn401WhenGettingUserByIdUnauthorised() {
```
ZaryabK: Does this test name/assertion need to change to reflect a 403 instead of a 401?

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -108,6 +116,50 @@ public void shouldThrowErrorWhenNewEmailExists() {
         userService.editUser(token.getToken(), userId, update);
     }
 
+    @Test(expected = UserForbiddenException.class)
+    public void shouldThrowErrorWhenRequestingManagerDoesNotMatchRequestedUser() {
+        given(authServiceImpl.getTokenById(token.getToken())).willReturn(token);
```
ZaryabK: Most of these givens are shared between tests, so they could be extracted into the setup method to help readability

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
MerrettM: No this actually does throw a 401, the unit tests check the throwing of the 403 errors.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -118,7 +115,16 @@ private void checkIfManagerExists(Long managerId) {
...
     private void checkTokenMatchesUser(UUID token, Long userId) {
         if (!authServiceImpl.getTokenById(token).getUserId().equals(userId)) {
-            throw new UserUnauthorisedException();
+            throw new UserForbiddenException();
+        }
+    }
+
+    private void checkUserRequestingIsAuthorized(Long requestedUserId, UUID token) {
+        User requestingUser = getUserById(authServiceImpl.getTokenById(token).getUserId());
+        User requestedUser = getUserById(requestedUserId);
```
ed-steele-rdm-els: I'd suggest doing this in MySQL - you can merge this because I assume it works how you'd expect, but sequential queries like this scale poorly. you could try e.g. 

```

SELECT EXISTS (
    SELECT 1
    FROM users
    WHERE manager_id = :requestingUserId
    OR id = :requestingUserManagerId
);
```
which will return as a Boolean that you can throw your exception based on this

I don't understand `checkTokenMatchesUser(token, requestedUserId);` it looks like you are checking the user whose details are being requested has logged in to your app - presumably these tokens expire so what happens if this user hasn't logged in in a while? This might be me failing to understand your architecture ... 

Btw to get your user from an auth token on 1 trip to the db (you want to do this) try 

```
SELECT (your,cols,etc)
FROM users usrs
INNER JOIN tokens tkns ON (usrs.id = tkns.user_id)
WHERE tkns.token = :tokenIncoming
```

You might have heard some stuff about JOINS being expensive - in this case joining 2 columns with autoincrementing INTs will be very quick - think about how you might find an element in an ordered list/array vs unordered in Java to understand why this might be quick 

(you don't need to explictly write the `INNER` but I'm including it to make this comment make more sense)

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -108,6 +116,50 @@ public void shouldThrowErrorWhenNewEmailExists() {
         userService.editUser(token.getToken(), userId, update);
     }
 
+    @Test(expected = UserForbiddenException.class)
```
ed-steele-rdm-els, line 119: Use the code coverage tool in intelliJ to ensure you've covered all branches with these tests (e.g. run all tests with coverage)

ed-steele-rdm-els: then I'll approve - heavy handed hint but I'm not sure you're covering case 2 of your if statement

H-Ford (approved):

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -54,4 +54,21 @@ void updateUser(@Param("id") Long id,
             "ORDER BY last_name ASC",
             nativeQuery = true)
     List<User> searchByName(@Param("query") String query);
+
+    @Query(value = "SELECT EXISTS(" +
```
ed-steele-rdm-els, line 58: Nice work;

One small thing 
`manager_id IS NOT NULL` is not needed, MySql is not Java, this will simply result in false if your Long is null

ed-steele-rdm-els (approved):

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
Piotrek605, line 58: Would it work with just SELECT 1 and the method returning a boolean as @ed-steele-rdm-els suggested?

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
```diff
@@ -54,4 +54,21 @@ void updateUser(@Param("id") Long id,
...
+            "(manager_id IS NOT NULL " +
+            "AND manager_id = :requestingUserId) " +
+            "OR " +
+            "(:requestingUserManagerId IS NOT NULL " +
+            "AND id = :requestingUserManagerId)" +
+            "))",
+            nativeQuery = true)
+    int hasManagerEmployeeRelationship(@Param("requestedUserId") Long requestedUserId,
+                                       @Param("requestingUserId") Long requestingUserId,
+                                       @Param("requestingUserManagerId") Long requestingUserManagerId);
```
Piotrek605, line 73: The manager ID of the requesting user can be inferred from the table, so it's not needed as a parameter?

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
MerrettM: Ah cool, I wasn't sure after doing some research so I used that just in case. Thanks!

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -118,7 +115,15 @@ private void checkIfManagerExists(Long managerId) {
...
     private void checkTokenMatchesUser(UUID token, Long userId) {
         if (!authServiceImpl.getTokenById(token).getUserId().equals(userId)) {
-            throw new UserUnauthorisedException();
+            throw new UserForbiddenException();
+        }
+    }
+
+    private void checkUserRequestingIsAuthorized(Long requestedUserId, UUID token) {
+        User requestingUser = getUserById(authServiceImpl.getTokenById(token).getUserId());
+        getUserById(requestedUserId);
```
Piotrek605, line 124: Is this still needed if the check is now done in the SQL statement?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -139,15 +143,59 @@ public void shouldReturn200WhenGettingUserById() {
...
+    @Test
+    public void shouldReturn401WhenGettingUserByIdUnauthorisedDueToToken() {
         ResponseEntity<ErrorEntity> response = getUserByIdExpectingAuthError();
 
         Assert.assertEquals(401, response.getStatusCodeValue());
         Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
     }
 
+    @Test
+    public void shouldReturn403WhenGettingUserWhoIsNotTheUserOrTheirManagerOrTheirEmployee() throws URISyntaxException {
```
Piotrek605, line 188: This sounds really confusing, how about just shouldReturn403WhenGettingUserForbidden? Imagine you change that logic later, you'll have to also change the name of this test.

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
ed-steele-rdm-els: It would, but the point of using EXISTS is that the the transaction can end as soon as one row matches (i.e. doesn't have to iterate whole filtered table) SELECT 1 returns 1 for each row that matches, on an app this size it makes no difference but I always prefer `EXISTS` - which would return 1/0 which would need to be mapped to a boolean in either case.

I think returning an int is OK for now (depending on the complexity of binding a result set mapper here)

MerrettM: Yeah I did try with returning a boolean but It wasn't having it so I decided to go with the int method. I can put it in a ticket as a fix to do after the feature freeze though?

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
MerrettM: Is it not easier to just put it in as a parameter rather than having to retrieve it from the requesting user's row?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
MerrettM: Well the second line checks the user you're requesting exists and throws a 404 if not whereas the if statement logic throws a 403 so yes I think?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
MerrettM: True, I didn't really like the name when I wrote it, this does read better.

SpringProject/src/main/java/com/gradproject2019/users/repository/UserRepository.java
Piotrek605: I suppose it's easier to write, it's up to you.

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
Piotrek605: You're right, they throw different exceptions

PR #27, opened 2019-12-06
Gp19 100 show interested endpoint

GBurley-Jones:


SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -77,11 +77,12 @@ private void checkNotInPast(Instant dateTime) {
             if (!dateTime.isAfter(Instant.now())) {
                 throw new InvalidConferenceFieldException();
             }
-        } catch (NullPointerException ignored) {}
+        } catch (NullPointerException ignored) {
```
Piotrek605: Why would you ignore NullPointerException? :O

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.UUID;
+
+@CrossOrigin(origins = "http://localhost:3000")
+@Controller
+public class UserConferenceController {
+
+    @Autowired
+    private UserConferenceService userConferenceService;
+
+    @PostMapping(path = "/user-conferences")
```
Piotrek605, line 24: Has this been approved by a senior helper?

SpringProject/src/main/java/com/gradproject2019/userConference/persistance/UserConferenceId.java
```diff
@@ -0,0 +1,36 @@
+package com.gradproject2019.userConference.persistance;
```
Piotrek605: persistEnce

SpringProject/src/main/java/com/gradproject2019/userConference/persistance/UserConferenceId.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+import java.io.Serializable;
+
+
+public class UserConferenceId implements Serializable {
```
Piotrek605: What's the difference between this and UserConference?

Piotrek605 (changes requested):

SpringProject/src/main/java/com/gradproject2019/userConference/persistance/UserConferenceId.java
GBurley-Jones: I believe this file is required because we are using a composite primary key, so this defines what is contained in the key. Before we were not using this and the project would not compile so it does appear to make a difference.

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
GBurley-Jones: This is an old part of the code base so I am not sure exactly who has written it, however I think it is because we are throwing a custom error instead.

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
GBurley-Jones: I am not exactly sure what you are referring to here? we discussed the Post mapping path with Harry

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
```diff
@@ -0,0 +1,85 @@
...
+import static com.gradproject2019.userConference.data.UserConferenceRequestDto.UserConferenceRequestDtoBuilder.anUserConferenceRequestDto;
+import static org.springframework.http.HttpMethod.POST;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class UserConferenceControllerIT extends TestUtils {
+    @LocalServerPort
+    int testServerPort;
+
+    private String baseUri;
```
H-Ford: Could just change this to a URI instead of a string

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public UserConferenceResponseDto saveInterest(UUID token, UserConferenceRequestDto userConferenceRequestDto) {
+        UserConference userConference = new UserConferenceRequestDto().from(userConferenceRequestDto);
+
+        checkUserAuthorised(token);
+
+        try {
+            UserConference savedUserConference = userConferenceRepository.saveAndFlush(userConference);
+            return new UserConferenceResponseDto().from(savedUserConference);
+        } catch (DuplicateKeyException e) {
```
H-Ford, line 36: You should check whether the user is already interested before saving the conference, in a separate method

GBurley-Jones: I think that because we are using a composite key the same user cannot register interest in a conference twice, this is why we have included a catch for the DuplicateKeyException which will be thrown if the user does try to. Do we also need a method to check this if the functionality is already there?

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
GBurley-Jones: If we use a URI here we cannot concatenate the baseUri to include the testServerPort, is there another method for doing this?

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
Piotrek605: It might still be useful for you to refactor this in this PR or at least find an explanation why it's written that way, up to you. NullPointerException is thrown when you're trying to use an object that doesn't exist and it's a bad idea to just ignore it.

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
Piotrek605: Fair enough, just bear in mind Harry isn't the only helper you should be learning from.

SpringProject/src/main/java/com/gradproject2019/userConference/persistance/UserConferenceId.java
Piotrek605: Would you mind explaining this to me in person?

Piotrek605 (approved):

H-Ford (approved):

PR #30, opened 2019-12-16
Gp19 101 get user conference endpoint

GBurley-Jones:


SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
```diff
@@ -54,14 +57,22 @@ public void shouldReturn200AndSaveInterestInConference() {
...
         Assert.assertEquals(401, response.getStatusCodeValue());
 
     }
+    @Test
+    public void shouldReturn200AndListOfConferencesUserIsInterestedIn() throws URISyntaxException {
+        URI uri = new URI (baseUri + "/" + savedConference.getId());
+
+        ResponseEntity<List<ConferenceResponseDto>> response = getConferenceByUserId(uri);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
```
ZaryabK, line 90: Should you be asserting on more than the status code here?

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -90,9 +96,17 @@ private void checkConferenceExists(Long conferenceId) {
...
     private void checkUserAuthorised(UUID token) {
         try {
-            authServiceImpl.checkTokenExists(token);
+            authService.checkTokenExists(token);
         } catch (TokenNotFoundException e) {
             throw new UserUnauthorisedException();
         }
     }
+
+    private void checkIfUserIsInterested(Long conferenceId) {
```
ZaryabK: What is this method doing at the moment?

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
```diff
@@ -38,11 +45,32 @@ public UserConferenceResponseDto saveInterest(UUID token, UserConferenceRequestD
...
         try {
             authService.checkTokenExists(token);
         } catch (TokenNotFoundException e) {
             throw new UserUnauthorisedException();
         }
     }
+
+    @Override
+    public boolean checkUserIsInterestedInConference(Long conferenceId){
+        if (userConferenceRepository.findByConferenceId(conferenceId).size() == 0) {
```
ZaryabK: This method could probably be better served by a "doesInterestExist" method in your repository which would return a boolean if a conference does exist :)

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
```diff
@@ -38,11 +45,32 @@ public UserConferenceResponseDto saveInterest(UUID token, UserConferenceRequestD
         }
     }
 
+    @Override
+    public List<ConferenceResponseDto> getConferenceByUserId(UUID token, Long userId) {
+        List<ConferenceResponseDto> conferenceResponseDtos = new ArrayList<>();
+
+        checkUserAuthorised(token);
```
ZaryabK, line 50: A minor point but should the token authorisation be happening in this method? I would imagine it would be better to check this in the API before this method is even called

ZaryabK (changes requested):

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
GBurley-Jones: Yes, this point has been raised a couple of times. There is a ticket in the backlog for changing the structure of this for all of our authorisation. But this has not been pulled into this weeks sprint as we are focusing on getting as many features as possible done before making these improvements.

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
GBurley-Jones: This is a good point,  as this method is actually more relevant to the ticket for de-expressing interest (and is reliant on the delete endpoint) I am going to remove it from this PR.

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
GBurley-Jones: Nothing, it was intended as a bare bones for checking if a user is interested in a conference before we delete it however this is more relevant to a different ticket.

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
```diff
@@ -38,11 +45,32 @@ public UserConferenceResponseDto saveInterest(UUID token, UserConferenceRequestD
         }
     }
 
+    @Override
+    public List<ConferenceResponseDto> getConferenceByUserId(UUID token, Long userId) {
+        List<ConferenceResponseDto> conferenceResponseDtos = new ArrayList<>();
```
Piotrek605: There's no point creating the list if the user isn't authorized, so this could go after the next line,

Piotrek605 (changes requested):

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
ed-steele-rdm-els: Also what does `if <condition> then false else true` tell you about the condition?
You can just return `!<condition>`

SpringProject/src/main/java/com/gradproject2019/userConference/service/UserConferenceServiceImpl.java
```diff
@@ -38,11 +45,32 @@ public UserConferenceResponseDto saveInterest(UUID token, UserConferenceRequestD
...
+
+        checkUserAuthorised(token);
+        List<UserConference> userConferences = userConferenceRepository.findByUserId(userId);
+
+        for(UserConference userConference: userConferences) {
+            conferenceResponseDtos.add(conferenceService.getConferenceById(userConference.getConferenceId()));
```
ed-steele-rdm-els, line 55: So this isn't a critism, working code is great at this stage, but be aware that this is inefficient code - this makes a db call for each element in this list. 

Changing this is out of scope of the project, but you'd want to select the specific columns you are returning to the FE, by joining between tables. Ask if you have any questions about this 👍

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
```diff
@@ -27,4 +25,10 @@
         UserConferenceResponseDto newInterest = userConferenceService.saveInterest(token, userConferenceRequestDto);
         return ResponseEntity.ok(newInterest);
     }
+
+    @GetMapping(path = "/user-conferences/{id}")
```
ed-steele-rdm-els, line 29: So there is a slight issue with this - without documentation any FE wouldn't know what the ID is referring to. Is it the Conference or the User? 
As you're developing the FE/BE so closely together and its working for you don't change it, but a path like `user/{id}/conferences` would make more sense to me that this returns a list of conferences associated to a user

SpringProject/src/main/java/com/gradproject2019/userConference/repository/UserConferenceRepository.java
```diff
@@ -1,10 +1,29 @@
...
 import org.springframework.stereotype.Repository;
 
+import java.util.List;
+import java.util.Optional;
+
 @Repository
 public interface UserConferenceRepository extends JpaRepository<UserConference, Long> {
 
+    @Query(value =
+            "SELECT * " +
```
ed-steele-rdm-els: Fine here, try not to write SELECT * in production code though (it can make very big result sets if joining to many tables)

ed-steele-rdm-els (approved):

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
GBurley-Jones: Ah I see what you are saying, I will discuss with FE and see if they would like me to rewrite this

SpringProject/src/main/java/com/gradproject2019/userConference/repository/UserConferenceRepository.java
GBurley-Jones: Okay thank you, Is there a better query for selecting all data of a certain type?

ed-steele-rdm-els: Yes - your 'dto's are effectively a map between the data in a column, and the variable in Java - so you would be need only select the columns that go into the Java Object

ed-steele-rdm-els: > selecting all data of a certain type 

do you mean like selecting all integers or something? That's not really possible

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
ZaryabK: It should probably be added in the other PR then, else you'll have a bunch of fluff in your code base :)

SpringProject/src/main/java/com/gradproject2019/userConference/repository/UserConferenceRepository.java
GBurley-Jones: So is SELECT * WHERE not okay? I thought that this was just selecting the columns are you are saying

ed-steele-rdm-els: Yes its evil 
https://use-the-index-luke.com/blog/2013-08/its-not-about-the-star-stupid
https://softwareengineering.stackexchange.com/a/234678

You should get out of the habit of using it in application code. In this case it _is_ just selecting the columns you need in this case because there are only 2 columns in the table AFAIK and they are both mapped to your dao. 

This isn't about this PR really, its more for your future when you write code in real systems

GBurley-Jones: Okay thanks Ed, I will give those a read :)

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
```diff
@@ -24,11 +26,12 @@
 
     private ConferenceRepository conferenceRepository;
 
-    private AuthServiceImpl authServiceImpl;
+    private AuthService authService;
```
ZaryabK, line 29: Curious as to why this has changed from AuthServiceImpl to AuthService?

SpringProject/src/test/java/com/gradproject2019/UserConferenceServiceTest.java
```diff
@@ -48,4 +50,18 @@ public void shouldSaveInterestInConference() {
         assertThat(userConference.getUserId()).isEqualTo(userConferenceResponseDto.getUserId());
 
     }
+
+    @Test
+    public void shouldGetInterestInConference() {
```
ZaryabK: I think you could probably add a few more test cases here, this seems like a good happy path test but there are a few unhappy paths that are untested

SpringProject/src/test/java/com/gradproject2019/UserConferenceServiceTest.java
```diff
@@ -48,4 +50,18 @@ public void shouldSaveInterestInConference() {
...
+        UserConference userConference = anUserConference()
+                .withUserId(1L)
+                .withConferenceId(1L)
+                .build();
+        given(userConferenceRepository.saveAndFlush(any(UserConference.class))).willReturn(userConference);
+
+        List<ConferenceResponseDto> conferenceResponseDtoList = userConferenceService.getConferenceByUserId(token.getToken(), userConference.getUserId());
+
+        assertThat(userConference.getConferenceId()).isEqualTo(conferenceResponseDtoList.get(0).getId());
```
ZaryabK: Because you won't know the size of the list at this point, maybe you should assert that there's only one element here

ZaryabK (changes requested):

Piotrek605 (approved):

SpringProject/src/main/java/com/gradproject2019/conferences/service/ConferenceServiceImpl.java
GBurley-Jones: Harry pointed out that we should always be referring to the Service layer through its interface and not the implementation

Piotrek605: I agree, otherwise what's the point having an interface and an implementation. The implementation might be replaced with another one at any point.

PR #31, opened 2019-12-19
Gp19 131 get users by manager user ID

karenzamani:


SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -43,4 +43,10 @@
         UserResponseDto userResponseDto = userService.findUserById(userId, token);
         return ResponseEntity.ok(userResponseDto);
     }
-}
+
+    @GetMapping(path = "/users/employee/{id}") //managers user ID
```
ed-steele-rdm-els: The idea of having a fixed parameter (i.e. {id}) in a path like this is that it uniquely identifies some kind of entity. In this case, what you are doing really is filtering all the users by their manager's id. 

In this path, it looks as if that id refers to a type of a `user` called an `employee` that is identified by an {id}. I suggest having this as a query param (i.e. `?manager_id={id}) of the base list users endpoint, that does the same thing. Having a comment to explain this doesn't help anyone using your API. 

This may seem like a minor point, but in an Application Programming _Interface_ the endpoints are all your clients see of your app - getting them right is the most important part of being a good back end dev.

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -176,6 +180,36 @@ public void shouldReturn200WhenGettingEmployeeById() throws URISyntaxException {
         Assert.assertEquals(employee.getOccupation(), response.getBody().getOccupation());
     }
 
+    @Test
+    public void shouldReturn200WhenGettingEmployeesByUserId() {
```
Piotrek605: This should say 'getting employees by manager id'

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -176,6 +180,36 @@ public void shouldReturn200WhenGettingEmployeeById() throws URISyntaxException {
...
+    public void shouldReturnEmptyListWhenUserHasNoEmployees() {
+
+        ResponseEntity<List<UserResponseDto>> response = getUserByManagerIdNoneManager(uri4);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(0, response.getBody().size());
+    }
+
     @Test
     public void shouldReturn401WhenGettingUserByIdUnauthorisedDueToToken() {
```
Piotrek605: I think a bad token is the only way to be unauthorized, so 'due to token' is redundant here.

Piotrek605 (approved):

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -64,6 +64,14 @@ public UserResponseDto editUser(UUID token, Long userId, UserPatchRequestDto use
         return new UserResponseDto().from(getUserById(userId));
     }
 
+    @Override
+    public List<UserResponseDto> findUserByManagerId(UUID token, Long managerId) {
+        checkTokenMatchesUser(token, managerId);
```
ed-steele-rdm-els, line 70: Not quite: it is absolutely necessary in your framework to have a dao that does just this filter, but what you want here is a general `getUsersOp` that can choose and control the right database query to run. 

So for an intial cut of this, you might have a line of code that checks managerId is not null and is a positive integer, and if it is run the filter users by manager id, if not return an unfiltered list of users. 

Subquequently you'll want to choose the filters based on iterating the query params as a map, and include paging - _but not right now_!

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -38,13 +37,18 @@
...
+    private URI uri3;
+    private URI uri4;
 
     @Before
     public void setUp() throws URISyntaxException {
         universalSetUp();
         baseUri = "http://localhost:" + testServerPort + "/users";
         uri = new URI(baseUri);
         uri2 = new URI(baseUri + "/" + savedUser.getId());
+        uri3 = new URI(baseUri  +"?manager_id=" + savedManager.getId());
```
ed-steele-rdm-els: I'm pretty sure you can dynmically build queries using the spring testing framework

ed-steele-rdm-els (changes requested):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -38,13 +37,18 @@
...
+        uri4 = new URI(baseUri + "?manager_id=" + savedUser.getId());
```
ZaryabK: 1. Do you need this and `uri3`?
2. uri1, uri2, uri3 etc. are not very clean variable names as it doesn't tell you anything about the URI itself - can you change these to be more descriptive?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -58,7 +62,7 @@ public void shouldReturn200AndSaveUserInDatabase() {
         HttpEntity<UserRequestDto> request = new HttpEntity<>(userRequestDto);
 
         ResponseEntity<String> response = postUser(request);
-        User retrievedUser = userRepository.findAll().get(1);
+        User retrievedUser = userRepository.findAll().get(3);
```
ZaryabK, line 65: 3 seems like a bit of a magic number here - what significance does it have to this test?

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -176,6 +180,36 @@ public void shouldReturn200WhenGettingEmployeeById() throws URISyntaxException {
...
     }
 
+    @Test
+    public void shouldReturn200WhenGettingEmployeesByUserId() {
+        ResponseEntity<List<UserResponseDto>> response = getUserByManagerId(uri3);
+
+        Assert.assertEquals(200, response.getStatusCodeValue());
+        Assert.assertEquals(response.getBody().size(), 1);
+        Assert.assertEquals(savedUserWithManager.getId(), response.getBody().get(0).getId());
+        Assert.assertEquals(savedUserWithManager.getFirstName(), response.getBody().get(0).getFirstName());
```
ZaryabK: I would argue these extra assertions don't add any value, just checking the ID is the same should be fine (correct me if I'm wrong)

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
```diff
@@ -305,7 +342,23 @@ public void shouldReturnUsersMatchingFullName() throws URISyntaxException {
...
+        return restTemplate.exchange(uri, GET, null, new ParameterizedTypeReference<List<UserResponseDto>>() {
+        });
+    }
+
+    private ResponseEntity<List<UserResponseDto>> getUserByManagerId(URI uri) {
+        return restTemplate.exchange(uri, GET, new HttpEntity<>(managerPassingHeaders), new ParameterizedTypeReference<List<UserResponseDto>>() {
+        });
+    }
+
+    private ResponseEntity<List<UserResponseDto>> getUserByManagerIdNoneManager(URI uri) {
```
ZaryabK: `NoneManager`?

SpringProject/src/it/java/com/gradproject2019/utils/TestUtils.java
```diff
@@ -32,13 +36,21 @@
...
+    public User userWithManager;
+    public User manager;
     public Conference savedConference;
     private String hashedPassword;
     public Token testToken;
     public HttpHeaders passingHeaders;
+    public HttpHeaders managerPassingHeaders;
     public HttpHeaders failingHeaders;
+    public User savedUserWithManager;
+    public User savedManager;
```
ZaryabK, line 57: Is there a way we can collapse some of these fields? i.e. can `manager` and `savedManager` be the same user? can `user1` or `user2` assume the other roles?

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
```diff
@@ -64,6 +64,14 @@ public UserResponseDto editUser(UUID token, Long userId, UserPatchRequestDto use
         return new UserResponseDto().from(getUserById(userId));
     }
 
+    @Override
+    public List<UserResponseDto> findUserByManagerId(UUID token, Long managerId) {
```
ZaryabK: Minor point but as this is returning a list of users, probably `findUsers` would be a better prefix

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -36,12 +38,18 @@
...
+    private final User user4 = new User(4L, "qwerty4", "qwerty4", "qwerty4", "qwerty4@qwerty.com", "Qwerty!4", "qwerty4", 3L);
+    private final User user5 = new User(5L, "qwerty5", "qwerty5", "qwerty5", "qwerty5@qwerty.com", "Qwerty!5", "qwerty5", 3L);
+
     private final Token token = new Token(1L, UUID.randomUUID());
+    private final Token managerToken = new Token(3L, UUID.randomUUID());
+
     private final Long userId = 1L;
     private final Long userId2 = 2L;
-
+    private final Long managerId = 3L;
```
ZaryabK: You could declare this above the users block above, and use it when creating your User objects

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
```diff
@@ -36,12 +38,18 @@
     @Mock
     private AuthServiceImpl authServiceImpl;
 
-    private final User qwerty = new User( 1L, "qwerty", "qwerty", "qwerty", "qwerty@qwerty.com", "Qwerty!1", "qwerty", null);
-    private final User qwerty2 = new User( 2L, "qwerty2", "qwerty2", "qwerty2", "qwerty2@qwerty.com", "Qwerty!1", "qwerty2", 3L);
+    private final User user = new User(1L, "qwerty", "qwerty", "qwerty", "qwerty@qwerty.com", "Qwerty!1", "qwerty", null);
+    private final User user2 = new User(2L, "qwerty2", "qwerty2", "qwerty2", "qwerty2@qwerty.com", "Qwerty!1", "qwerty2", 3L);
+    private final User manager = new User(3L, "qwerty3", "qwerty3", "qwerty3", "qwerty3@qwerty.com", "Qwerty!3", "qwerty3", null);
+    private final User user4 = new User(4L, "qwerty4", "qwerty4", "qwerty4", "qwerty4@qwerty.com", "Qwerty!4", "qwerty4", 3L);
+    private final User user5 = new User(5L, "qwerty5", "qwerty5", "qwerty5", "qwerty5@qwerty.com", "Qwerty!5", "qwerty5", 3L);
```
ZaryabK: This seems quite messy to me, which is a typical sign of a code smell.
1. Do you have a builder for users? If so you should be using that instead
2. Do the other fields (besides userId and managerId) need to be here? If not then you could give them all the same default values when using the builder

ZaryabK (changes requested):

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: I'll make a new ticket to update our testing as we've got some other tests to update as well!

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: Yep :)

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: it's just because i created 2 extra users for testing so now there are 3 in total. There is probably a better way to test this that won't break if new users are added

SpringProject/src/main/java/com/gradproject2019/users/service/UserServiceImpl.java
karenzamani: changed :)
i like the minor points, they are easier to rectify haha

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: haha you told me to add extra assertions!!
tbf i only had the 200 assertion at the time so if id is all that needed to be added i'll remove the others

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: should be when user is not a manger

SpringProject/src/test/java/com/gradproject2019/UserServiceTest.java
karenzamani: i don't need user 4 and user 5, i just wanted to check a list would be returned but i will remove them

SpringProject/src/it/java/com/gradproject2019/user/UserControllerIT.java
karenzamani: removed addition users so this does not apply any more!

SpringProject/src/main/java/com/gradproject2019/users/controller/UserController.java
```diff
@@ -43,4 +43,10 @@
         UserResponseDto userResponseDto = userService.findUserById(userId, token);
         return ResponseEntity.ok(userResponseDto);
     }
-}
+
+    @GetMapping(path = "/users")
```
ed-steele-rdm-els, line 47: OK, Looking at this again I realise I've been a darn fool 
There is no need to pass in the `manager_id` at all - you have that from the `token` (indirectly).

So in your business you should use the mapping/token table to retrieve the user/manager id, then build the query based on that.  

So the base query for get users should be;

```
SELECT * FROM users 
WHERE manager_id=:userId
OR user_id = :userId
```

If I understand it right the only users you can see when using the app, are the current user's employees? (and yourself). 

If you introduce admins/superusers you would need to remove this default filter

Then additional query params, like search name would just add new filters to that query.

ed-steele-rdm-els (approved):
Merge away

PR #32, opened 2019-12-19
Gp19 145 delete user conference

MerrettM:
Users should be able to de-express interest in a conference.

(also fixes bug where deleting conference would fail if a user had favourited it)

### DIFF
--- a/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
@@ -60,7 +60,6 @@ public class UserConferenceControllerIT extends TestUtils {
         ResponseEntity<UserConferenceResponseDto> response = saveInterestExpectingAuthError(baseUri, userConferenceRequestDto);
 
         Assert.assertEquals(401, response.getStatusCodeValue());
-
     }
 
     private ResponseEntity<UserConferenceResponseDto> saveInterest(URI uri, UserConferenceRequestDto request) {
--- a/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
@@ -138,7 +138,7 @@ public class UserConferenceControllerIT extends TestUtils {
 
     @Test
     public void shouldReturn404WhenUserConferenceNotFound() throws URISyntaxException {
-        URI uri = new URI(baseUri.toString() + "/2");
+        URI uri = new URI(baseUri.toString() + "/" + Integer.MAX_VALUE);
 
         ResponseEntity<ErrorEntity> response = deleteUserConferencesExpectingError(uri);
 
--- a/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
@@ -1,8 +1,10 @@
 package com.gradproject2019.userConference;
 
+import com.gradproject2019.conferences.data.ConferenceResponseDto;
 import com.gradproject2019.userConference.data.UserConferenceRequestDto;
 import com.gradproject2019.userConference.data.UserConferenceResponseDto;
 import com.gradproject2019.userConference.persistence.UserConference;
+import com.gradproject2019.utils.ErrorEntity;
 import com.gradproject2019.utils.TestUtils;
 import org.junit.After;
 import org.junit.Assert;
@@ -18,22 +20,29 @@ import org.springframework.test.context.junit4.SpringRunner;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.List;
 
-import static com.gradproject2019.userConference.data.UserConferenceRequestDto.UserConferenceRequestDtoBuilder.anUserConferenceRequestDto;
-import static org.springframework.http.HttpMethod.POST;
+import static org.springframework.http.HttpMethod.*;
 
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 public class UserConferenceControllerIT extends TestUtils {
+
     @LocalServerPort
     int testServerPort;
 
     private URI baseUri;
+    private URI baseUserUri;
+    private URI baseDeleteUri;
+    private UserConferenceRequestDto userConferenceRequestDto;
 
     @Before
     public void setUp() throws URISyntaxException {
         universalSetUp();
         baseUri = new URI("http://localhost:" + testServerPort + "/user-conferences");
+        baseUserUri = new URI(baseUri.toString() + "/" + savedUser.getId());
+        baseDeleteUri = new URI(baseUserUri.toString() + "/" + savedConference.getId());
+        userConferenceRequestDto = new UserConferenceRequestDto(savedUser.getId(), savedConference.getId());
     }
 
     @After
@@ -43,9 +52,8 @@ public class UserConferenceControllerIT extends TestUtils {
 
     @Test
     public void shouldReturn200AndSaveInterestInConference() {
-        UserConferenceRequestDto userConferenceRequestDto = createRequestDto(savedUser.getId(), savedConference.getId());
-
-        ResponseEntity<UserConferenceResponseDto> response = saveInterest(baseUri, userConferenceRequestDto);
+        userConferenceRepository.deleteAll();
+        ResponseEntity<UserConferenceResponseDto> response = saveInterest(userConferenceRequestDto);
         UserConference retrievedUserConference = userConferenceRepository.findAll().get(0);
 
         Assert.assertEquals(200, response.getStatusCodeValue());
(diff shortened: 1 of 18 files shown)


SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
```diff
@@ -26,9 +26,15 @@
         return ResponseEntity.ok(newInterest);
     }
 
-    @GetMapping(path = "/user-conferences/{id}")
-    public ResponseEntity<List<ConferenceResponseDto>> getConferenceByUserId(@RequestHeader("Authorization") UUID token, @PathVariable("id") Long userId) {
+    @GetMapping(path = "/user-conferences/{userId}")
```
Piotrek605: I should've said that in the previous PR but userId here also acts as a filter, so it should be a query parameter rather than a path one, am I right Ed?

SpringProject/src/test/java/com/gradproject2019/ConferenceServiceTest.java
```diff
@@ -23,7 +26,7 @@
 public class ConferenceServiceTest {
 
     @InjectMocks
-    private ConferenceServiceImpl conferenceService;
+    private ConferenceServiceImpl conferenceServiceImpl;
```
Piotrek605, line 26: This should be of type 'ConferenceService', shouldn't it?

SpringProject/src/test/java/com/gradproject2019/UserConferenceServiceTest.java
```diff
@@ -1,75 +1,72 @@
...
 
     @Mock
     private UserConferenceRepository userConferenceRepository;
 
+    @Mock
+    private AuthServiceImpl authServiceImpl;
 
     @Mock
-    private ConferenceService conferenceService;
+    private ConferenceServiceImpl conferenceServiceImpl;
```
Piotrek605, line 41: Same here

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
```diff
@@ -26,9 +26,15 @@
...
+    public ResponseEntity<List<ConferenceResponseDto>> getConferenceByUserId(@RequestHeader("Authorization") UUID token, @PathVariable("userId") Long userId) {
         List<ConferenceResponseDto> conferences = userConferenceService.getConferenceByUserId(token, userId);
         return ResponseEntity.ok(conferences);
     }
+
+    @DeleteMapping(path = "/user-conferences/{userId}/{conferenceId}")
```
Piotrek605: Here you access a resource by its key so it would be fine to have it as a path parameter, but it's a composite key and there's no way you can put those parameters next to each other. Doing it this way is wrong because you're not accessing the resource first by userId and then by conferenceId. Let's see what other helpers suggest.

Piotrek605 (changes requested):

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
MerrettM: I had a look online and this seemed to be the suggested method, I wasn't sure it looked right though so if we do it differently I can do that

SpringProject/src/test/java/com/gradproject2019/ConferenceServiceTest.java
MerrettM: when I make it a `ConferenceService` I get this error: `Cannot instantiate @InjectMocks field named 'conferenceService'! Cause: the type 'ConferenceService' is an interface.`

SpringProject/src/main/java/com/gradproject2019/userConference/controller/UserConferenceController.java
tessahoad: I think that a nicer path would be `GET users/{user_id}/conferences` - this also then solves the other problem for the DELETE path - we can have `DELETE users/{user_id}/conferences/{conference_id}`. 

There is a slight problem with this suggestion - that is that we are assuming the only relationship between users and conference is "interested in", but maybe that's a problem to worry about later

ed-steele-rdm-els: Yeah - I agree. The issue with the `user-conferences` is it makes it sound like an entity (when really its just an link between 2 different entities (user/conference).

I suggested having a path of `users/{id}/conferences`, but a query param of `/users` to filter conferences is equally fine. 

I think whoevers running this scheme should make sure we all understand what the difference between entities and database tables is - i.e. an entity is a business object with some intrinsic meaning, a database is a way of structing data. In this case of relational databases, e.g. MySql it structures the _relations_ between entities. Entities are the things which go in your paths, db tables are not. 

In this case a row in a db is definitely not an entity because it has no intinsic meaning (its just 2 numbers), and it just a _relation_ between users and conferences. 

I feel that forcing the writing java beans whenever you need to interact with a database has confused the gosh darn blazes out of everyone.

MerrettM: So we just realised that we actually don't need a `{userId}` as a path variable because users should only be able to access/edit/delete/etc. their own data so we can just use the `userId` field we retrieve from the database when we check their `UUID` is valid. this means we can actually just have the get and post endpoints as `/user-conferences` and the delete as `/user-conferences/{conferenceId}`.

Piotrek605 (approved):

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
```diff
@@ -47,9 +50,8 @@ public void tearDown() {
 
     @Test
     public void shouldReturn200AndSaveInterestInConference() {
-        UserConferenceRequestDto userConferenceRequestDto = userConferenceCreateRequestDto(savedUser.getId(), savedConference.getId());
-
-        ResponseEntity<UserConferenceResponseDto> response = saveInterest(baseUri, userConferenceRequestDto);
+        userConferenceRepository.deleteAll();
```
ZaryabK, line 53: You should be deleting before and after every test in the set up methods - is this explicit call to delete the records using the repository necessary here?

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
```diff
@@ -59,75 +61,120 @@ public void shouldReturn200AndSaveInterestInConference() {
...
-        ResponseEntity<UserConferenceResponseDto> response = saveInterestExpectingAuthError(baseUri, userConferenceRequestDto);
+        ResponseEntity<ErrorEntity> response = saveInterestExpectingAuthError(userConferenceRequestDto);
 
         Assert.assertEquals(401, response.getStatusCodeValue());
+        Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
+    }
 
+    @Test
+    public void shouldReturn404AndNotSaveWhenConferenceDoesNotExist() {
+        userConferenceRepository.deleteAll();
```
ZaryabK, line 72: Similar to a previous comment, does this need to be here?

SpringProject/src/test/java/com/gradproject2019/ConferenceServiceTest.java
```diff
@@ -46,14 +46,14 @@ public void shouldGetListOfAllConferences() {
...
+        ConferenceResponseDto conferenceById = conferenceServiceImpl.getConferenceById(1L);
 
         assertThat(conferenceById.getId()).isEqualTo(1L);
         assertThat(conferenceById.getName()).isEqualTo(conference1.getName());
     }
 
     @Test(expected = ConferenceNotFoundException.class)
     public void shouldThrowErrorWhenIdNotRecognised() {
-        conferenceService.getConferenceById(1000000000L);
+        conferenceServiceImpl.getConferenceById(1000000000L);
```
ZaryabK: Can I suggest Long.MAX_VALUE here?

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
```diff
@@ -59,75 +61,120 @@ public void shouldReturn200AndSaveInterestInConference() {
...
+    public void shouldReturn401WhenUserRequestingDeleteNotLoggedIn() {
+        ResponseEntity<ErrorEntity> response = deleteUserConferencesExpectingAuthError();
 
         Assert.assertEquals(401, response.getStatusCodeValue());
+        Assert.assertEquals("User unauthorized to perform action.", response.getBody().getMessage());
+    }
+
+    @Test
+    public void shouldReturn404WhenUserConferenceNotFound() throws URISyntaxException {
+        URI uri = new URI(baseUri.toString() + "/2");
```
ZaryabK: Is there a reason that a conference with an ID of 2 doesn't exist?

If you're looking for a number that clearly won't ever be used, maybe INTEGER.MAX_VALUE would be better here.

ZaryabK (changes requested):

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
MerrettM: Yes, this method requires an empty user conference repository, and while the universal setup method does delete from all repositories it also the populates them with standard data which is used in almost all tests, just not this one.

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
MerrettM: Ditto to the previous comment.

SpringProject/src/it/java/com/gradproject2019/userConference/UserConferenceControllerIT.java
MerrettM: ah yeah, my bad

PR #34, opened 2020-01-06
Gp19 125 assert on error messages

MerrettM:
Adds an assert for all error messages in integration tests.

### DIFF
--- a/SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
+++ b/SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
@@ -15,6 +15,7 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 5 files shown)


SpringProject/src/it/java/com/gradproject2019/conferences/ConferenceControllerIT.java
```diff
@@ -188,23 +193,39 @@ public void shouldReturn200AndEditOnlyNotNullFields() throws URISyntaxException
...
     private ResponseEntity<String> deleteConference(URI uri) {
         return restTemplate.exchange(uri, DELETE, new HttpEntity<>(passingHeaders), new ParameterizedTypeReference<String>() {
         });
     }
 
+    private ResponseEntity<ErrorEntity> deleteConferenceExpectingError(URI uri) {
+        return restTemplate.exchange(uri, DELETE, new HttpEntity<>(passingHeaders), ErrorEntity.class);
+    }
+
+    private ResponseEntity<ErrorEntity> deleteConferenceExpectingAuthError(URI uri) {
```
ELS-Zaryab: Small note - this seems very similar to the method above it but for the headers that are either passing or failing - they could be combined into 1 method with an extra param 'headers'? (only if you think it will improve readability in the tests)

ELS-Zaryab (approved):

H-Ford (approved):

PR #35, opened 2020-01-06
Gp19 elasticsearch write

GBurley-Jones:
In this branch I have moved the main application into a new maven module so this is all the same code as previously. Only the Elasticsearch code needs reviewing


SpringProject/ElasticSearch/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '2.0'
+
+services:
+  mysql:
+    image: mysql:5.7
```
ed-steele-rdm-els: Why are you using an older version of mysql? 8.0 is the latest version

ed-steele-rdm-els (commented):
I can't really review the functionality of this (there's a lot to unpack), but I left some vague style guideline comments

SpringProject/ElasticSearch/docker-compose.yml
GBurley-Jones: Good question, I all honesty I am not sure there is a reason! I have updated this to say mysql:8.0 and it still runs so I guess its okay?

ed-steele-rdm-els:
> I can't really review the functionality of this (there's a lot to unpack), but I left some vague style guideline comments

Which github seems to have lost... 
basically; 

- var https://openjdk.java.net/projects/amber/LVTIstyle.html - don't overuse it. You should be writing the type of an object unless there is some good reason not to, e.g. `var someList = new ArrayList<String>();` 
- If created or deleting your indexes fail then the logs don't help find which index failed, and they'll still exist - look into having a retry policy or giving more info in the logs
- There's a bit when you load this json file based on where it is in the project - any reason for not putting that info into more standard config?

SpringProject/ElasticSearch/docker-compose.yml
ed-steele-rdm-els: Yes probably although I suspect Mendeley infra can't handle 8.0 so check where this is being built and deployed (if it is), if you're building everything from stratch it should be fine

GBurley-Jones: We are building from scratch so I think it should be fine with 8.0 thanks!

Piotrek605 (approved):

PR #37, opened 2020-01-07
GP19-129: remove shared paths to top of each controller

MerrettM:


ELS-Zaryab:
Not sure where this comment would go, but personally I prefer the way it is now

MerrettM:
> Not sure where this comment would go, but personally I prefer the way it is now

This was suggested by Manol as a minor change to cut down on repeated code, as a group the consensus seems to be that we don't really mind either way but its done now so... `¯\_(ツ)_/¯`

Piotrek605 (approved):

